Issue #12, opened 2021-05-02
Add UTCI stress categories


grjonathan:
Could models.utci also return the UTCI stress category (e.g. strong heat stress)?  There is a table with the ranges in [Zare et al. (2018)](https://doi.org/10.1016/j.wace.2018.01.004).

FedericoTartarini:
Hi, this can surely be done, however, I have had a look at the [UTCI official tool](http://www.utci.org/index.php) but the tool does not return the same categories as shown in Table 2. See screenshot below.

![image](https://user-images.githubusercontent.com/40018640/116970248-35362e00-acea-11eb-9ce9-bd60cda2541c.png)

grjonathan:
You are correct, it is not a return from the official tool. However, if you look in the [documents section](http://www.utci.org/utci_doku.php) of that same page you will see the categories in "UTCI Assessment Scale.doc". 
I believe the scale originally appeared in [this paper by  Blazejczyk et al.](https://www.researchgate.net/publication/239525964_An_introduction_to_the_Universal_Thermal_Climate_Index_UTCI)

![image](https://user-images.githubusercontent.com/51000403/117087286-fbcae600-ad1c-11eb-86a1-bfc73eaaeb11.png)

FedericoTartarini:
Thank you a lot for providing the code and helping me with this. 

I have released a new version of pythermalcomfort which now includes your code. I have only made one small change and renamed one of the key (`stress_category`) that is returned by the UTCI function.

grjonathan:
Great, thanks for adding it!

PR #13, opened 2021-05-02
UTCI stress categories

grjonathan:
Issue #12
This adds the UTCI stress category (a string) to the output of models.utci.
Alternatively, this could be split into a separate function so that the output of models.utci remains a float.

### DIFF
--- a/src/pythermalcomfort/models.py
+++ b/src/pythermalcomfort/models.py
@@ -1096,6 +1096,8 @@ def utci(tdb, tr, v, rh, units="SI"):
     -------
     utci : float
          Universal Thermal Climate Index, [°C] or in [°F]
+    Stress Category : str
+         UTCI categorized in terms of thermal stress.
 
     Notes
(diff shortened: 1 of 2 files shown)


FedericoTartarini:
Thank you for working on this but we need to make sure we do not break the code for people who are already using the UTCI old code. We could either: 
* add a warning stating that the UTCI code has changed 
* pass an input like `return_stress_categories` which is False by default. 
* create an additional function in utilities which calculates the stress categories.

grjonathan:
Thank you for having a look. I agree this feature shouldn't break existing applications nor the examples/tutorials. A separate function in utilities might make sense, but there would have to be a toggle for units. If users are piping outputs directly from models.utci, they might forget to set the units a second time.

I added a `return_stress_category` input set to False by default and reverted the test. Please let me know what you think.

FedericoTartarini (approved):

Issue #2, opened 2020-05-11
Calculate MRT from climate model output?


milankl:
Hey @FedericoTartarini, great package!

I'm looking for a way to calculate UTCI from [ERA5](https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-land?tab=overview) data. To do that, I need to get the mean radiant temperature from the radiation fluxes provided by ERA5, available are

-  Forecast albedo
- Surface latent heat flux
- Surface net solar radiation
- Surface net thermal radiation
- Surface sensible heat flux
- Surface solar radiation downwards
- Surface thermal radiation downwards

Do you have experience how to do that, maybe already written some functions?

FedericoTartarini:
Hi, I am very happy you find the package useful. 

Unfortunately, I have not done it before, however, you can start by looking at the [ERF function](https://pythermalcomfort.readthedocs.io/en/latest/reference/pythermalcomfort.html#solar-gain-on-people) which estimates the effects of solar radiation on human heat gain and comfort indoors. [Link to the paper](https://escholarship.org/uc/item/89m1h2dg).

There may also be some other papers already published on this topic. Kindly share your findings with me if possible, so we may add a function to this package, if need be.

milankl:
Yeah, I haven't really worked on this before... For the UTCI do you use the polynomial approximation? Apparently, there are three different ways to calculate it, with different speeds & accuracies as shown in Fig. 2 of [this poster](http://utci.org/utci_poster.pdf)

FedericoTartarini:
Chris Mackey originally wrote the UTCI equation in Python and he used the code that it is available [here](http://www.utci.org)

milankl:
You mean the .f90 implementation somewhere on that website? Yes, that uses the polynomial approximation.

FedericoTartarini:
Yes, I refer to the .f90 file.

milankl:
This [dataset](https://cds.climate.copernicus.eu/cdsapp#!/dataset/derived-utci-historical?tab=overview) already includes the MRT for my purpose (and contains UTCI as well). I will contact the authors on details of the calculation, and let you know in case you want to implement them into your package.

milankl:
Okay this calculation is described [here](https://link.springer.com/content/pdf/10.1007/s00484-020-01900-5.pdf) in case you want to add it. It requires a handful of radiative fluxes and location and time of day for some angles, which I guess is already part of some other functions.

[...]

PR #24, opened 2022-03-10
utci numpy

### LINKED ISSUE #23: numpy api for utci
Hey everybody,
we would be interested in using some thermal comfort indicators on gridded datasets, e.g., `numpy` arrays. I am not sure how you think about this. In general it only requires minor changes, e.g., replacing `if` statements with `np.where` and returning `np.nan` instead of raising errors or warning. I have implemented the `utci` index as an example (i can show in a pull request) and the changes in the code are really minor. I think other indicators could be adapted too and they would still work on single floats but also on whole numpy arrays....

larsbuntemeyer:
I would like to have comfort indicators work on numpy array. Here i start with `utci`:

to do:

- [x] replace `if` statements with `np.where`
- [x] removed `ValueError` and warnings for invalid input ranges, return `np.nan` instead.
- [x] replaced stress category `if` statements with `np.digitize` binning.
- [x] add examples
- [ ] update docs
- [x] add tests
- [ ] add test for invalid input
- [ ] add example notebook

It all should still work with floats but now also with numpy arrays (and much faster than using `np.vectorize`)

closes #23

### DIFF
--- a/src/pythermalcomfort/models.py
+++ b/src/pythermalcomfort/models.py
@@ -1105,35 +1105,13 @@ def utci(tdb, tr, v, rh, units="SI", return_stress_category=False):
 
     utci_approx = np.where(all_valid, utci_approx, np.nan)
 
-    return np.round_(utci_approx, 1)
-    #if utci_approx < -40:
-    #    stress_category = "extreme cold stress"
-    #elif utci_approx < -27:
-    #    stress_category = "very strong cold stress"
-    #elif utci_approx < -13:
-    #    stress_category = "strong cold stress"
-    #elif utci_approx < 0:
-    #    stress_category = "moderate cold stress"
-    #elif utci_approx < 9:
-    #    stress_category = "slight cold stress"
-    #elif utci_approx < 26:
-    #    stress_category = "no thermal stress"
-    #elif utci_approx < 32:
-    #    stress_category = "moderate heat stress"
-    #elif utci_approx < 38:
-    #    stress_category = "strong heat stress"
-    #elif utci_approx < 46:
-    #    stress_category = "very strong heat stress"
-    #else:
-    #    stress_category = "extreme heat stress"
+    if units.lower() == "ip":
+        utci_approx = units_converter(tmp=utci_approx, from_units="si")[0]
 
-    #if units.lower() == "ip":
-    #    utci_approx = units_converter(tmp=utci_approx, from_units="si")[0]
-
-    #if return_stress_category:
-    #    return {"utci": round(utci_approx, 1), "stress_category": stress_category}
-    #else:
-    #    return round(utci_approx, 1)
+    if return_stress_category:
+        return {"utci": np.round_(utci_approx, 1), "stress_category": map_stress_category(utci_approx)}
+    else:
+        return np.round_(utci_approx, 1)
 
 
 def clo_tout(tout, units="SI"):
@@ -2417,6 +2395,25 @@ def use_fans_morris(
 
     return tipping_point
 
+
+def map_stress_category(t):
+    """Maps a temperature array to stress categories."""
+
+    thresholds = {-40.:'extreme cold stress',
+                   -27.:'very strong cold stress',
+                   -13.:'strong cold stress',
+                     0.: 'moderate cold stress',
+                    9.: 'slight cold stress',
+                    26.: 'no thermal stress', 
+                    32.: 'moderate heat stress',
+                    38.: 'strong heat stress',
+                    46.: 'very strong heat stress',
+                   1000.: 'extreme heat stress'}
+    
+    bins = np.array(list(thresholds.keys()))
+    words = np.append(np.array(list(thresholds.values())), 'unknown')
+    return words[np.digitize(t, bins, right=True)]
+    
 
 # # testing morris equation
 # from scipy import optimize
--- a/src/pythermalcomfort/models.py
+++ b/src/pythermalcomfort/models.py
@@ -1018,13 +1018,13 @@ def utci(tdb, tr, v, rh, units="SI", return_stress_category=False):
 
     Parameters
     ----------
-    tdb : float
+    tdb : float, array_like
         dry bulb air temperature, default in [°C] in [°F] if `units` = 'IP'
-    tr : float
+    tr : float, array_like
         mean radiant temperature, default in [°C] in [°F] if `units` = 'IP'
-    v : float
+    v : float, array_like
         wind speed 10m above ground level, default in [m/s] in [fps] if `units` = 'IP'
-    rh : float
+    rh : float, array_like
         relative humidity, [%]
     units: str default="SI"
         select the SI (International System of Units) or the IP (Imperial Units) system.
@@ -1033,9 +1033,9 @@ def utci(tdb, tr, v, rh, units="SI", return_stress_category=False):
 
     Returns
     -------
-    utci : float
+    utci : float, array_like
          Universal Thermal Climate Index, [°C] or in [°F]
-    stress_category : str
+    stress_category : str, array_like
          UTCI categorized in terms of thermal stress [9]_.
 
     Notes
--- a/src/pythermalcomfort/models.py
+++ b/src/pythermalcomfort/models.py
@@ -1071,7 +1071,7 @@ def utci(tdb, tr, v, rh, units="SI", return_stress_category=False):
     if units.lower() == "ip":
         tdb, tr, v = units_converter(tdb=tdb, tr=tr, v=v)
 
-    #check_standard_compliance(standard="utci", tdb=tdb, tr=tr, v=v)
+    # check_standard_compliance(standard="utci", tdb=tdb, tr=tr, v=v)
 
     def exponential(t_db):
         g = [
@@ -1092,9 +1092,9 @@ def utci(tdb, tr, v, rh, units="SI", return_stress_category=False):
 
     # Do a series of checks to be sure that the input values are within the bounds
     # accepted by the model.
-    tdb_valid = valid_range(tdb, (-50.,50.))
-    diff_valid = valid_range(tr - tdb, (-30.,70.))
-    v_valid = valid_range(v, (0.5, 17.))
+    tdb_valid = valid_range(tdb, (-50.0, 50.0))
+    diff_valid = valid_range(tr - tdb, (-30.0, 70.0))
+    v_valid = valid_range(v, (0.5, 17.0))
     all_valid = ~(np.isnan(tdb_valid) | np.isnan(diff_valid) | np.isnan(v_valid))
 
     eh_pa = exponential(tdb) * (rh / 100.0)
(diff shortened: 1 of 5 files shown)


larsbuntemeyer:
i am sure why the tests fail, seems unrelated to these changes here. can i run the travis ci on the pull request?

src/pythermalcomfort/models.py
```diff
@@ -1069,7 +1071,7 @@ def utci(tdb, tr, v, rh, units="SI", return_stress_category=False):
     if units.lower() == "ip":
         tdb, tr, v = units_converter(tdb=tdb, tr=tr, v=v)
 
-    check_standard_compliance(standard="utci", tdb=tdb, tr=tr, v=v)
```
FedericoTartarini, line 1072: I have noticed that you removed this check. How can we return a warning if the inputs are outside the UTCI model limits of applicability?

src/pythermalcomfort/models.py
```diff
@@ -1082,60 +1084,56 @@ def exponential(t_db):
...
-    elif utci_approx < 32:
-        stress_category = "moderate heat stress"
-    elif utci_approx < 38:
-        stress_category = "strong heat stress"
-    elif utci_approx < 46:
-        stress_category = "very strong heat stress"
-    else:
-        stress_category = "extreme heat stress"
-
-    if units.lower() == "ip":
```
FedericoTartarini, line 1132: Have you removed this function? We need to convert back the results into IP units if the user selected them.

src/pythermalcomfort/models.py
```diff
@@ -1105,35 +1105,13 @@ def exponential(t_db):
...
-    #    stress_category = "extreme heat stress"
-
-    #if units.lower() == "ip":
-    #    utci_approx = units_converter(tmp=utci_approx, from_units="si")[0]
-
-    #if return_stress_category:
-    #    return {"utci": round(utci_approx, 1), "stress_category": stress_category}
-    #else:
-    #    return round(utci_approx, 1)
+    if units.lower() == "ip":
```
FedericoTartarini, line 1108: Should not we allow this function to get a numpy array as an input?

src/pythermalcomfort/models.py
```diff
@@ -2420,6 +2399,27 @@ def use_fans_morris(
     return tipping_point
 
 
+def map_stress_category(t):
```
FedericoTartarini, line 2402: please move this function to `utilities.py`. If possible I want to keep the `models.py` file clean and include in it only the functions that calculate the thermal comfort indices.

tests/test_models.py
```diff
@@ -1345,17 +1346,19 @@ def test_ankle_draft():
...
-            {"tdb": 25, "tr": 27, "rh": 50, "v": 1, "return": {"utci": 25.2}},
-            {"tdb": 19, "tr": 24, "rh": 50, "v": 1, "return": {"utci": 20.0}},
-            {"tdb": 19, "tr": 14, "rh": 50, "v": 1, "return": {"utci": 16.8}},
-            {"tdb": 27, "tr": 22, "rh": 50, "v": 1, "return": {"utci": 25.5}},
-            {"tdb": 27, "tr": 22, "rh": 50, "v": 10, "return": {"utci": 20.0}},
-            {"tdb": 27, "tr": 22, "rh": 50, "v": 16, "return": {"utci": 15.8}},
-        ]
-    )
+@pytest.fixture
+def data_test_adaptive_ashrae():
```
FedericoTartarini, line 1350: could you please change this from `data_test_adaptive_ashrae` to `data_test_utci`. I originally erroneously named this variable.

FedericoTartarini (commented):
Thank you for the great suggestions and for allowing users to now pass arrays as input to the function. I am not sure what it is causing the issue. After you reply to my minor comments I will Pull the changes and try to understand which tests are failing.

src/pythermalcomfort/models.py
larsbuntemeyer: I noticed that the standard compliance test for `utci` here:
https://github.com/CenterForTheBuiltEnvironment/pythermalcomfort/blob/fc143fe26637765518d97dcdd93d9fb821d56d89/src/pythermalcomfort/utilities.py#L27-L33
raises a warning for wind speed. However, i noticed that this was checked again anyway:
https://github.com/CenterForTheBuiltEnvironment/pythermalcomfort/blob/fc143fe26637765518d97dcdd93d9fb821d56d89/src/pythermalcomfort/models.py#L1093-L1100

The warning is just a fundamental difference if we have arrays here, i guess? If i have a large array, e.g., a large geospatial dataset like ERA5 as suggested here #2 , there might be valid and invalid input combination at the same time and you don't want to throw millions of warnings as you would have for a geospatial dataset. So instead, i suggest to return `np.nan` instead of raising a warning or the `ValueError`.

Long story short: I am not sure, if we would break anything for users if the function returns `np.nan` instead of raising a `ValueError` (although most of the time going on with nan should also raise an error...) For now, i could introduce a check, if the input is just a single value to still raise an error and a warning that this might change in the future, what do you think?

src/pythermalcomfort/models.py
larsbuntemeyer: `units_converter` already does accect numpy arrays, since it's just basic math snytax supported by numpy!

src/pythermalcomfort/models.py
larsbuntemeyer: i noticed that there was no test for the case a users input was invalid....

src/pythermalcomfort/models.py
larsbuntemeyer: no problem!

larsbuntemeyer:
thanks for looking at this @FedericoTartarini !

src/pythermalcomfort/models.py
FedericoTartarini: One example of how returning `nan` instead of raising a warning may be a problem is presented below. 

The ASHRAE and ISO have slightly different applicability limits for the PMV model. If we return a warning and we still calculate a PMV value nothing will break in our users' code. If on the other hand, we return a `nan` then their code will break. I know that most people ignore warnings and returning a `nan` is more effective and informative, but at the same time, for example, users will find it strange that for example for a `tdb = 49.9` we return a UTCI value while for  `tdb = 50.1` we do not.

What are your thoughts on this?

FedericoTartarini: You are right there was no test. I did not implement all the necessary tests. More tests are welcome :)

FedericoTartarini: A few options on how to solve this issue are below:
* we return `nan` only if the user inputs arrays, while we raise a warning if the user inputs a single value and we clearly document that.
* we calculate a UTCI value even if the input is outside the model limit, but we also return an array with boolean values listing for each element if the inputs exceeded the applicability limits.

Perhaps the second solution is a bit better. I am open to other suggestions.

larsbuntemeyer: i would also prefer the second solution, maybe even in the form of an argument like `ignore_limits`, so if that's true, also values outside of the applicability limits will be returned, otherwise `np.nan`. By default, it should be `False`.

larsbuntemeyer: > One example of how returning `nan` instead of raising a warning may be a problem is presented below.
> 
> The ASHRAE and ISO have slightly different applicability limits for the PMV model. If we return a warning and we still calculate a PMV value nothing will break in our users' code. If on the other hand, we return a `nan` then their code will break. I know that most people ignore warnings and returning a `nan` is more effective and informative, but at the same time, for example, users will find it strange that for example for a `tdb = 49.9` we return a UTCI value while for `tdb = 50.1` we do not.
> 
> What are your thoughts on this?

at least, for `utci` also for invalid float input, there would be an error, so returning `np.nan` instead would make sure, that user's code still breakes if they didn't take care of the `ValueError` before...

FedericoTartarini: > i would also prefer the second solution, maybe even in the form of an argument like `ignore_limits`, so if that's true, also values outside of the applicability limits will be returned, otherwise `np.nan`. By default, it should be `False`.

I agree, let's then use the second solution and add an input argument `ignore_limits`. Would you be able to implement the changes?

larsbuntemeyer: yes, no problem!

FedericoTartarini: Let me know once you are done so I can accept your pull request and release a new version of pythermalcomfort. Thank you again for working on this. You are the first person from the community who contributes to the package :)

FedericoTartarini (approved):

FedericoTartarini:
Thank you so much for working on this and for your help. If you have time I would like to implement the same also for the other models such as the PMV, SET,... since this is a great feature that you have implemented. Please let me know if you would have a bit of spare time to work on it. I will pull the changes locally, check that everything is fine and then I will release a new version of pythermalcomfort.

larsbuntemeyer:
Yes, i will check out the other models as soon as possible. Mostly, it's does not require many change if you have some experience with numpy! I will also add the notebook to the examples!

FedericoTartarini:
Great, yes please add the notebook to the examples since it would be super valuable to other users. If you want I can also do it. Let me try to implement what you have done for the UTCI inside the `pmv_ppd` function so I learn something new and I will then get back to you if I face some issues implementing it.

Thank you again for your work. It is now included inside the new pythermalcomfort version. Any other feedback you have about the tool would be highly appreciated.

FedericoTartarini:
I have been working on converting the PMV and SET function to accept a NumPy array as input but I experienced some issues.

Inside the PMV and SET functions there are a lot of while and if loops.
I converted the if -> np.where as you did, but I do not know how to handle the while loops ([for example](https://github.com/CenterForTheBuiltEnvironment/pythermalcomfort/blob/75248885e019430804709fc718874d56a1c9c78b/src/pythermalcomfort/optimized_functions.py#L371)) and this seems to be causing some issues. Do you happen to know how to solve the problem?

Shall I use [nditer](https://numpy.org/doc/stable/reference/arrays.nditer.html)

larsbuntemeyer:
> I have been working on converting the PMV and SET function to accept a NumPy array as input but I experienced some issues.
> 
> Inside the PMV and SET functions there are a lot of while and if loops. I converted the if -> np.where as you did, but I do not know how to handle the while loops ([for example](https://github.com/CenterForTheBuiltEnvironment/pythermalcomfort/blob/75248885e019430804709fc718874d56a1c9c78b/src/pythermalcomfort/optimized_functions.py#L371)) and this seems to be causing some issues. Do you happen to know how to solve the problem?

Is that a kind of iterative solver (`pmv_ppd_optimized`)? I'll check this out, maybe there is some clever way of doing vectorized iterations in numpy...
> 
> Shall I use [nditer](https://numpy.org/doc/stable/reference/arrays.nditer.html)

I think, for a start, you could check out [`np.vectorize`](https://numpy.org/doc/stable/reference/generated/numpy.vectorize.html), it's a simple way of applying a function on all array elements. However, this can become slow, if the function itself is not vectorized...

> 
> Moreover, rather than having to re-write the whole code, would it not be better to just check as soon as the function is called if the user passed an array as input. If that is the case, then we just call the existing functions passing a single element rather than an array. What do you think?

That could probably be an intermediate solution? I just think that could produce some code duplication.


Should we make this a new issue?

FedericoTartarini:
`pmv_ppd_otimized` is just a normal python function that is wrapped with Numba to significantly improve its performances. 

At this stage, speed is not a major concern since all the functions are already optimized with Numba. Yes, both `np.vectorize` and `np.nditer` should both work fine. In addition, if we use this solution we do not need to change the functions themselves and we can keep them more readable. What do you mean with "if the function is not vectorized?"

I would not want to duplicate any functions since that would make the code very difficult to maintain. 

Yes, we can open a new issue. We could also look into how other packages, such as `scikit learn` are handling vectors as inputs.

PR #59, opened 2023-05-15
docs: fixing JOS3 documentation

AkihisaNomoto:
Hi, I have modified the documentation of JOS3 and also updated the test results of the model. It is small modification so I tried to change the master blanch of _pythermalcomfort/CenterForTheBuiltEnvironment_, but my puch failed for some reason. Therefore, I opened this pull request.


src/pythermalcomfort/jos3_functions/parameters.py
```diff
@@ -115,7 +115,7 @@
         "suffix": "Body name",
         "unit": "W",
     },
-    "met_base_fat": {
+    "Q_bmr_fat": {
```
FedericoTartarini, line 118: would it be possible to avoid using capital letters in the name of the variables? I have noticed that some of them are capitalised and some are not, it would be better to have them all in this format `xx_yy_zz` (we may not need to have two dashed, it was just to show the formatting)

src/pythermalcomfort/models.py
```diff
@@ -3432,6 +3434,7 @@ def __init__(
         weight_loss_by_evap_and_res: weight loss by the evaporation and respiration of the whole body [g/sec]
         OPTIONAL PARAMETERS : the paramters listed below are returned if ex_output = "all"
         Q_bmr_core: core heat production by basal metabolism (each body part) [W]
```
FedericoTartarini, line 3436: This is referred to the previous comment, as you can see all the Q are capitalised, it would be nice to have them as q_....

FedericoTartarini (commented):
I left a few minor comments. I blocked the master branch since in the future several people may work on this project and it is better that no one can push to it without prior opening a pull request. This is also the recommendation that GitHub gives.

src/pythermalcomfort/jos3_functions/parameters.py
AkihisaNomoto: I tried to separate heat gain and heat loss: I use _Q_ as heat production inside the body such as basal metabolism, work, non-shivering or shivering (heat gain) and _q_ as heat transfer (basically heat loss since body temperatures are higher than the ambient) such as convective heat transfer, radiant heat transfer and so on. I am not sure this is the right way to describe, and welcome to get the suggestion. What do you think?

src/pythermalcomfort/models.py
AkihisaNomoto: The same as above.

src/pythermalcomfort/jos3_functions/parameters.py
FedericoTartarini: now I see the point, however, since the q can also be an heat gain if we pass tdb > 35 then I would use all lower case so it is more pythonic, more consistent for the user, and it does not a priory assume heat gain or losses since they may vary. If you, however, you feel very strongly about keeping it as it is I am okay with that as well. It is just a small modification.

FedericoTartarini:
@AkihisaNomoto looking at the code again, I am wondering if we should replace dictionaries with @dataclasses. What worries me is that in some part of the code like line 4298 of models.py we access the units using this command `u = ALL_OUT_PARAMS[param]["unit"]`. This could be a big problem since the code does not really know that 'unit' is there we are just taking a leap of faith. If someone changes the ALL_OUT_PARAMS the the whole code will break.

```
ALL_OUT_PARAMS = {
    "age": {"ex_output": True, "meaning": "age", "suffix": None, "unit": "years"},
    "bf_ava_foot": {
....
```

If on the other had we use @dataclasses we can more programmatically access the values since we can enforce that the key `age` must be there and its argument should be another data class. Simple example below:

```
@dataclass()
class InfoParameters:
    ex_output: str
    meaning: str
    ....

@dataclass()
class AllOutputParameters:
    age: InfoParameters
```

I am new to data classes but this to me seems a much better approach so we can make sure the code will not break. What do you think?

src/pythermalcomfort/jos3_functions/parameters.py
AkihisaNomoto: I don't have strong preferences and tried to fix this, but I realized that we call total total heat loss from the skin (sensible and latent heat loss from the skin) as _q_skin_ which is the same variable name as other pythermalcomfort functions and call skin total heat production as _Q_skin_... So, I am thinking of just leaving this as it is.

AkihisaNomoto:
> @AkihisaNomoto looking at the code again, I am wondering if we should replace dictionaries with @dataclasses. What worries me is that in some part of the code like line 4298 of models.py we access the units using this command `u = ALL_OUT_PARAMS[param]["unit"]`. This could be a big problem since the code does not really know that 'unit' is there we are just taking a leap of faith. If someone changes the ALL_OUT_PARAMS the the whole code will break.

Sorry, I can not get the point. Could you please rephrase this? It seems natural to me that if someone changes the code, the program might stop working. Would @dataclasses make it more robust? I am not familiar with this, so let me research it.

src/pythermalcomfort/jos3_functions/parameters.py
FedericoTartarini: So basically we have the same name for two variables that represent two different heat loss/gains and the only difference is the capitalization? 

If this is the case I believe we should come up with unique names anyway otherwise this is very confusing. sensible and latent heat loss from the skin could be q_skil_lat_sens or q_skin_ev_cond while skin total heat production could be q_skin_gen. These are just suggestions but it is critical that we have two separate names otherwise other users and developers may struggle with this in the future. Relaying on capitalization alone is not the best strategy in my opinion.

FedericoTartarini:
yes, dataclasess are more robust and they will make the code much easier to maintain and we will avoid most of the errors in the future. I created a video to explain the benefits which I believe are huge and the change is very minor.

[This is a video explanation](https://unisyd-my.sharepoint.com/:v:/g/personal/federico_tartarini_sydney_edu_au/EeA2ccDODu9HvEAGzalwvBkBRcx19u4eT4l2gFel_MUUAQ?e=xSMXGm) of why dataclasses are great

src/pythermalcomfort/jos3_functions/parameters.py
AkihisaNomoto: Only the skin layer was in conflict because of heat exchange and heat production with the surrounding environment. For the other layers it is fine. 

As shown in the commit below, 
- Q_core, Q_muscle, and Q_skin were modified as q_thermogenesis_core, q_thermogenesis_muscle, and q_thermogenesis_skin. 
- Heat loss from the skin was changed to q_skin2env. 
- All symbols was modified to lowercase, and reorderd the output parametes of JOS3 model.

AkihisaNomoto:
Thanks for letting me know the dataclass. I opened a new issue so let's work on this part there. If you are fine with JOS3 documentation with my variable name change, please consider merging this pull request.

FedericoTartarini (approved):
@AkihisaNomoto I love your changes. Thank you so much for implementing them. I have merged the pull request

Issue #56, opened 2023-05-02
Write tests for JOS3 model


FedericoTartarini:
We should create some tests to validate the output of the JOS3 model. We can create a new file called `test_jos3.py` inside `tests` and then write inside there all the tests for the JOS3 functions.

AkihisaNomoto:
Do you want to test all the functions that JOS3 model uses? Let's discuss this later.

FedericoTartarini:
You can use ChatGPT to write some tests. They will not be perfect but it will save a lot of time so we can test all the functions.

My prompt to ChatGPT:

```
write some tests using pytest for this python function:

def bsa_rate(
    height=1.72,
    weight=74.43,
    formula="dubois",
):
    """Calculate the rate of bsa to standard body.

    Parameters
    ----------
    height : float, optional
        Body height [m]. The default is 1.72.
    weight : float, optional
        Body weight [kg]. The default is 74.43.
    formula : str, optional
        The equation name (str) of bsa calculation. Choose a name from "dubois",
        "takahira", "fujimoto", or "kurazumi". The default is "dubois".

    Returns
    -------
    bsa_rate : float
        The ratio of bsa to the standard body [-].
    """
    bsa_all = body_surface_area(
        height=height,
        weight=weight,
        formula=formula,
    )
    return bsa_all / _BSAst.sum()  # The bsa ratio to the standard body (1.87m2)
```

Its answer:
```
import pytest
from my_module import bsa_rate

def test_bsa_rate_with_defaults():
    assert bsa_rate() == pytest.approx(1.0)

def test_bsa_rate_with_custom_height_weight():
    assert bsa_rate(height=1.80, weight=80.0) == pytest.approx(1.1904, rel=1e-4)

def test_bsa_rate_with_invalid_formula():
    with pytest.raises(ValueError):
        bsa_rate(formula="invalid_formula_name")
```

Actually the last test would be very useful and now it is not working since this function `body_surface_area` has a bit of a problem since it does not return anything if the wrong model is passed. I have then fixed the `body_surface_area` function`.

I have pushed the changes so you can see the test I have implemented.

This [tool](https://app.codecov.io/github/CenterForTheBuiltEnvironment/pythermalcomfort) is very useful to see which functions of pythermalcomfort have been tested.

AkihisaNomoto:
[...]

PR #65, opened 2023-08-19
Test code for JOS-3 model (issue #56) and some updates

AkihisaNomoto:
I have fixed some of the coding bugs for JOS-3 model and have added test code (issue #56 ).

### Update for JOS-3 model
- changed the "t_skin" setter method for the user to allow to set custom values. I assume someone wants to set experimental values to JOS-3's skin temperature to set the initial condition of simulations.
- Based on [a reported issue on the original JOS3 repository](https://github.com/TanabeLab/JOS-3/issues/11), I have fixed the bugs. This change does not affect the simulation that much.

### Test code for JOS-3 model
- I have added and tested almost all test code for JOS-3 model in `tests/test_jos3.py`, and made some of the functions more robust, expecting to input invalid values. **Please take a look and let me know what you think.**
-  In `src/pythermalcomfort/jos3_functions/construction.py`, I added the default body information before calling the functions. This change allows us to change the default values more easily in the future.

### Others
I pushed some comfort model files that I am currently working on, which I should have done yet. So I have deleted the wrong push.

Aki


src/pythermalcomfort/jos3_functions/construction.py
```diff
@@ -38,6 +38,51 @@
     ]
 )
 
+# Anthropomorphic data for a standard body
+default_height = 1.72
```
FedericoTartarini, line 42: Great, I like this change since you only define these default values once in the code. On the other hand, should not we add them to a dataclass which contains all the defaults so it is easier to keep track of them and we do not have to prepend each variable name with the term default? Something like the following

```
@dataclass
class Defaults:
     height: float = 1.72
     weight: float = 74.43
```

src/pythermalcomfort/jos3_functions/thermoregulation.py
```diff
@@ -803,16 +808,16 @@ def ava_blood_flow(
         AVA blood flow rate at hand and foot [L/h].
     """
     # Cal. mean error body core temp.
-    cap_bcr = [10.2975, 9.3935, 13.834]  # Thermal capacity at chest, back and pelvis
+    cap_bcr = [10.2975, 9.3935, 4.488]  # Thermal capacity at chest, back and pelvis
```
FedericoTartarini, line 811: I trust you on these changes since I am not sure which one is the correct value. It would be nice to add a reference where you got the value from since aside from you only a few people in the world would be able to easily and quickly find the reference where you got these values from. The same is true from the change below in line 813.

src/pythermalcomfort/models.py
```diff
@@ -3724,7 +3725,7 @@ def _calculate_operative_temp_when_pmv_is_zero(
                 to = to - vpmv / 3
         return to
 
-    def _reset_setpt(self):
+    def _reset_setpt(self, par: float = 1.25):
```
FedericoTartarini, line 3728: Maybe we could expand a bit the documentation here since we need to make it clear why we are calling this function.

src/pythermalcomfort/models.py
```diff
@@ -3737,12 +3738,12 @@ def _reset_setpt(self):
...
         # par = 1.25
         # 1 met = 58.15 W/m2
-        met = self.bmr * 1.25 / 58.15  # [met]
+        met = self.bmr * par / 58.15  # [met]
         self.to = self._calculate_operative_temp_when_pmv_is_zero(met=met)
         self.rh = 50  # Relative humidity
         self.v = 0.1  # Air velocity
         self.clo = 0  # Clothing insulation
-        self.par = 1.25  # Physical activity ratio
+        self.par = par  # Physical activity ratio
```
FedericoTartarini, line 3746: Shall we rename `par` into `physical_activity_ratio` at first glance I thought that par meant parameter.

tests/test_jos3.py
```diff
@@ -1,15 +1,999 @@
...
+from pythermalcomfort.jos3_functions.construction import *
+from pythermalcomfort.jos3_functions.construction import _BSAst
+from pythermalcomfort.jos3_functions.matrix import *
+from pythermalcomfort.jos3_functions.thermoregulation import *
 
 
-def test_bsa_rate_with_defaults():
-    assert bsa_rate() == pytest.approx(1.00051943)
+"""
+Following test codes were basically generated by ChatGPT (Model:GPT-4) with some manual modification.
```
FedericoTartarini, line 12: No need to add this since it is not needed to know how you generated the code as long as it does the job. Less comments are better since we want to have a clean code.

tests/test_jos3.py
```diff
@@ -1,15 +1,999 @@
...
-def test_bsa_rate_with_invalid_formula():
+
+    # Test 2: Check the methods
+    # Instantiate the JOS3 class
+    model = JOS3()
+
+    # Call the simulate method
+    model.simulate(times=60)
+
+    # Test 2.1: _calculate_operative_temp_when_pmv_is_zero()
```
FedericoTartarini, line 35: Are the test numbers appearing in the console? If so it would be nice to keep them for debugging purposes, if on the other hand they are not appearing and are there in the code base only (Which I think that this is the case) I would remove the `Test XX` text since it can become confusing if we decide to change the order in which the tests re run. On the other hand a brief description of the test may be useful if the test is very complex or it comprises several steps.

tests/test_jos3.py
FedericoTartarini, line 1: It would also be great to add tests that make sure that the final results of a simulation are correct. We could think about two scenarios, for example:
* a person sitting at rest exposed to 25 and then 30C.
* a person walking exposed to 27C and 0.7m/s which then transition to an indoor space at 23C and 0.1m/s
The above are just ideas and more tests would be better, we just want to ensure that the final results will not change if someone changes any of the inner function. If of course in the future we will notice an error in the code we will then update these tests but at least we know and verify that the final output has changed.

Have you tested all the other functions you have included in JOS-3 or are there still some functions that have not been tested?

FedericoTartarini (commented):
Great work, thank you so much for your effort. I left a few comments in the code.

src/pythermalcomfort/jos3_functions/construction.py
AkihisaNomoto: OK. Created a default data class with anthropomorphic data for a default body in `construction.py`.

src/pythermalcomfort/jos3_functions/thermoregulation.py
AkihisaNomoto: This seems to be a typo. It was pointed out that the values do not match those shown in the paper, so it has been corrected.

tests/test_jos3.py
AkihisaNomoto: OK, removed this docstring.

FedericoTartarini:
@AkihisaNomoto I am getting a bit lost about what it is going on in this repo. I saw you responded to some of my comments but I cannot see the new code. In addition I can see that you merged your code into master.

image

AkihisaNomoto:
@FedericoTartarini Hi Federico - I synced CenterForTheBuiltEnvironment:master repo to my forked repo to update it to the latest version. I responded to some of your comments to clarify and note where I have finished modifying the issue. I committed those changes in my local but have not pushed them yet. Once I finish editing all the changes or at some point, I will push them. Sorry for the confusion!

FedericoTartarini:
Thank you. Sorry that I got confused. It all makes sense now. Thank you for working on this.

src/pythermalcomfort/models.py
AkihisaNomoto: I added the document.

src/pythermalcomfort/models.py
AkihisaNomoto: I think it can be left as is because this is a common abbreviation like PMV.

tests/test_jos3.py
AkihisaNomoto: No, I deleted the test numbers.

tests/test_jos3.py
AkihisaNomoto: I picked up the data in the example file, and now the test code for JOS3 class compares the predicted mean skin temperature with the one in the example file.  

I also added a human subject experiment dataset in the `jos3_output_example` folder to see the impact of future changes on the results predicted by JOS3. This is not necessary for the pythermalcmfort package, but it is useful for developers.

In the test file, all functions required for JOS3 calculations have been tested.

AkihisaNomoto:
I have updated the code, and please take a look and see if it is running on your end.

FedericoTartarini:
Hi @AkihisaNomoto could you please add `pandas` as a dependency in `tox` [here](https://github.com/CenterForTheBuiltEnvironment/pythermalcomfort/blob/fd1c9dcb960bf6f85928adc790a43370337b1125/tox.ini#L40). In the test you have implemented pandas is listed as a dependency, hence, we need to add it to the tox file otherwise the tests fail. Before pushing the code could you also please check that the tests all pass?

examples/calc_jos3.py
FedericoTartarini, line 1: Could you please reformat all the files with Black so the code style is consistent across the whole project?

src/pythermalcomfort/jos3_functions/construction.py
```diff
@@ -12,11 +12,24 @@
...
+     height: float = 1.72
+     weight: float = 74.43
+     age: int = 20
+     body_fat: float = 15
+     cardiac_index: float = 2.59  # [L/min/m2]
+     sex: str = "male"
+     posture: str = "standing"
+     bmr_equation: str = "harris-benedict"
+     bsa_equation: str = "dubois"
+     local_bsa: float = np.array( # body surface area [m2]
```
FedericoTartarini, line 32: I think you may have selected the wrong data type here since you are saying that local_bsa is a float while instead is an array of floats. Could you please check this?

src/pythermalcomfort/jos3_functions/construction.py
```diff
@@ -212,15 +280,15 @@ def bfb_rate(
     else:  # age >= 70
         ci *= 0.7
 
-    bfb_all = ci * bsa_rate(height, weight, bsa_equation) * _BSAst.sum()  # [L/h]
+    bfb_all = ci * bsa_rate(height, weight, bsa_equation) * Default.local_bsa.sum()  # [L/h]
     return bfb_all / 290
```
FedericoTartarini, line 284: Why are we dividing this term by 290? Could you please add a comment in which we explain what 290 is?

src/pythermalcomfort/jos3_functions/thermoregulation.py
```diff
@@ -15,33 +15,10 @@
...
-)
-
+from pythermalcomfort.jos3_functions.construction import Default
 
 def conv_coef(
-    posture="standing",
+    posture=Default.posture,
     v=0.1,
     tdb=28.8,
     t_skin=34.0,
```
FedericoTartarini, line 24: Should we add t_skin = 34 as a default parameter as well? The same can be done for the above temperature and wind speed event though V and T are less of a Default parameter.

examples/calc_jos3.py
AkihisaNomoto: OK, I have reformatted all the files related to JOS3 model with Black.

src/pythermalcomfort/jos3_functions/construction.py
AkihisaNomoto: Oh sorry, and thank you for pointing this out! I have changed it to np.ndarray from float.

src/pythermalcomfort/jos3_functions/construction.py
AkihisaNomoto: Thank you for pointing this out. This is really wrong way to code. 290 is the blood flow rate in the whole body. I added this to the Default class and used it in the function instead of using a numerical value.

src/pythermalcomfort/jos3_functions/thermoregulation.py
AkihisaNomoto: I added some other parameters to the default data class, such as t_skin, ta, tr, rh, etc and input those data to the functions instead of inputting the numerical values for us to easily change the code.

AkihisaNomoto:
Hi, @FedericoTartarini. Thank you for all your valuable comments. Those made the code better, especially using the Python data class really made it easy to code (I will change some parts to data class, but that might be a little later). I added pandas as a dependency in the tox file and checked that all the tests successfully passed.

src/pythermalcomfort/jos3_functions/thermoregulation.py
```diff
@@ -548,12 +541,15 @@ def evaporation(
...
     )  # bsa rate
-    bsa = _BSAst * bsar  # bsa
+    bsa = Default.local_bsa * bsar  # bsa
     p_a = antoine(tdb) * rh / 100  # Saturated vapor pressure of ambient [kPa]
     p_sk_s = antoine(t_skin)  # Saturated vapor pressure at the skin [kPa]
 
     e_max = (p_sk_s - p_a) / ret * bsa  # Maximum evaporative heat loss
 
+    # Replace any zero values in the e_max array with 0.001 to avoid causing a divide by 0 error
+    e_max[e_max == 0] = 0.001
```
FedericoTartarini, line 551: I am sure you have checked this code but I am not sure what e_max is (array or dataframe) so I want to check with you that this code works.

FedericoTartarini (commented):
Great work, I added two very minor comments and then we can merge the code into pythermalcomfort

src/pythermalcomfort/jos3_functions/thermoregulation.py
AkihisaNomoto: e_max is an array and means maximum evaporative heat loss from the skin at each body part. 

Here is the sample code:

```
import numpy as np

# Create an array named e_max (as sample data)
e_max = np.array([0, 1, 2, 0, 3, 0, 4])

# Replace any zero values in the e_max array with 0.001 to avoid causing a divide by 0 error
e_max[e_max == 0] = 0.001

# Output the modified e_max array
print(e_max)

```

Results

`[0.001 1.    2.    0.001 3.    0.001 4.   ]
`

We also have a test code in `test_jos3.py` to see if this code is working.

AkihisaNomoto: However, I modified it to the numpy.where method that many people probably use. It does the same thing.

src/pythermalcomfort/models.py
```diff
@@ -3739,12 +3745,12 @@ def _reset_setpt(self):
         # Set operative temperature under PMV=0 environment
         # par = 1.25
         # 1 met = 58.15 W/m2
-        met = self.bmr * 1.25 / 58.15  # [met]
+        met = self.bmr * par / 58.15  # [met]
```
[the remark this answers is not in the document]
AkihisaNomoto: Actually, metabolic rate in W/m2 is only used here. This is just doing a conversion from W/m2 to met. I added a new variable called `w_per_m2_to_met`  to convert the unit.

AkihisaNomoto:
@FedericoTartarini Thank you for your comment. I have changed them.

FedericoTartarini (approved):

PR #77, opened 2023-10-02
Initial Integration of Zhang’s comfort model and ATCS (Advanced Thermal Comfort Simulator)

AkihisaNomoto:
Hi @FedericoTartarini

- Added two new models in `models.py` file. One is Zhang Hui's comfort model from physiological parameters and the other is ATCS (Advanced Thermal Comfort Simulator).  Zhang model predicts thermal sensation and comfort from physiology.  ATCS is the integrated model of JOS3 and Zhang's model and it first predicts thermal physiology such as skin and core temperatures from the body built and ambient thermal environment, and then predicts thermal sensation and comfort from physiology predicted by JOS3. Zhang's model is described as a function and ATCS is a class because ATCS inherits JOS3 class, so the usage of ATCS is exactly the same as JOS3.
- Added relevant files such as `sensation.py`, `comfort.py` and so on. As we described before, the original overall sensation prediction is a little bit tricky so I wrote a very simple implementation for now. Ideally, I wanted to do a pull request once I perfectly implemented the model, but I will have a presentation in mid-Oct and will talk about this topic, so I would like you to merge this code for now. I will definitely work on modeling the original overall sensation prediction.
- Added example codes for both Zhang model and ATCS and these results. Zhang's model itself can be used to predict sensation and comfort when users have some experimental data on thermal physiology.


src/pythermalcomfort/zhang_comfort_model_functions/coefficients.py
FedericoTartarini, line 1: I am not sure what do to with these files, but my personal preference would be to have a single file about Hui's model which contains the code, coefficient and utilities. I think that splitting the code into separate files makes is much more complex to maintain. But this is a personal preference and I would love to hear what you think.

src/pythermalcomfort/zhang_comfort_model_functions/coefficients_summary.csv
FedericoTartarini, line 1: I think this file should be converted to another format and included within the code. Keeping it like this could be a big issue if we later decide to change a variable name. Here variables names are hardcoded and not linked to the code. See my previous comment where I am suggesting of combining all this files, related to one single model

src/pythermalcomfort/zhang_comfort_model_functions/utilities.py
```diff
@@ -0,0 +1,90 @@
+import numpy as np
+
+# 17 body segment names (The name corresponds to JOS-3 model)
+body_name_list = [
```
FedericoTartarini, line 4: shall we use dataclasses here? I am very worried of using strings for code maintainability in the future.

src/pythermalcomfort/zhang_comfort_model_functions/utilities.py
```diff
@@ -0,0 +1,90 @@
...
+        else:
+            ValueError("The input list or ndarray is not of length 17")
+    else:
+        raise ValueError(
+            "Unsupported input type. Supported types: int, float, list, dict, ndarray"
+        )
+    return data_17_array
+
+
+def convert_17_segments_dict_from_array(array, body_name_list):
```
FedericoTartarini, line 73: May be the case that this function, or a very similar function, is also used in JOS3 model?

src/pythermalcomfort/zhang_comfort_model_functions/utilities.py
FedericoTartarini, line 1: I do not think this should be a separate file. These functions should be moved to the main file which contains the model's code

src/pythermalcomfort/models.py
```diff
@@ -4633,3 +4643,676 @@ def bmr(self):
...
+
+            # Use the parent class's method to calculate the operative temperature where PMV is zero.
+            self.to = self._calculate_operative_temp_when_pmv_is_zero(
+                met=met, clo=overall_clo
+            )
+
+            # Run the model in uniform conditions at a temperature where PMV=0
+            # for the given met and clo and use the resulting skin temperatures as the setpoints
+            # Assuming that the initial conditions are uniform and at a temperature where PMV=0
+            print("Skin neutral temperature calculation starts")
```
FedericoTartarini, line 5174: should we consider using LOGs rather than prints?

src/pythermalcomfort/zhang_comfort_model_functions/coefficients.py
```diff
@@ -0,0 +1,479 @@
+import pandas as pd  # for making dataframe
+
+c1_plus = {
+    "head": 0.9,
```
FedericoTartarini, line 4: These should be data classes and not dictionaries since we have the same string repeated many times. You can create a dataclass and then use it to define `c1_plus`, `c1-minus`, ....

Moreover the variable name c1_plus is not very descriptive and we could improve it

src/pythermalcomfort/zhang_comfort_model_functions/coefficients.py
```diff
@@ -0,0 +1,479 @@
...
+    "right_hand": 0.0,
+    "left_thigh": 0.0,
+    "left_leg": 0.0,
+    "left_foot": 0.0,
+    "right_thigh": 0.0,
+    "right_leg": 0.0,
+    "right_foot": 0.0,
+}
+
+# Below is the code for making a summary table of coefficients
```
FedericoTartarini, line 422: removed this commented out code

src/pythermalcomfort/zhang_comfort_model_functions/comfort.py
```diff
@@ -0,0 +1,233 @@
...
+        overall_sensation : float
+            The overall thermal sensation which could be either positive or negative.
+
+        Returns
+        -------
+        tuple
+            A tuple containing the computed elements.
+        """
+        if overall_sensation >= 0:
+            A1 = -8 * (C6 + C72 * overall_sensation + 4)
```
FedericoTartarini, line 47: could you please explain what these functions are doing?

src/pythermalcomfort/zhang_comfort_model_functions/sensation.py
```diff
@@ -0,0 +1,244 @@
...
+    t_skin_local_set_array : numpy.ndarray
+        Array of local skin temperature setpoints [°C].
+    t_skin_mean_value : float
+        Mean skin temperature value [°C].
+    t_skin_mean_set_value : float
+        Setpoint for mean skin temperature [°C].
+    dt_core_local_dt_array : numpy.ndarray
+        Time derivative of core temperatures [°C/s].
+    sensation_scale : str
+        Thermal sensation scale to be used. Accepts either '9-point' or '7-point'.
```
FedericoTartarini, line 30: please specify which one is used as default

src/pythermalcomfort/zhang_comfort_model_functions/sensation.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def calculate_overall_sensation(sensation_local_array):
+    """Calculates overall sensation.
+
+    Parameters:
+    sensation_local_array (array): Array of sensations from various body parts.
+
+    :returns
+    overall_sensation_dict (dict): Dict ....
```
FedericoTartarini, line 110: could you please better describe the output?

src/pythermalcomfort/zhang_comfort_model_functions/sensation.py
```diff
@@ -0,0 +1,244 @@
...
+        # Extract the top two maximum and minimum sensations
+        max_two = sorted(sensation_local_array, reverse=True)[:2]
+        min_two = sorted(sensation_local_array)[:2]
+
+        # Set up hand and foot sensations
+        hand_sensations = {left_hand_sensation, right_hand_sensation}
+        foot_sensations = {left_foot_sensation, right_foot_sensation}
+
+        # Check if the most extreme sensations correspond to two hands or two feet
+        return (
```
FedericoTartarini, line 171: could you please reformat all the files with Black, the formatting here is very strange

FedericoTartarini (commented):
Great work, I left some comments. The major issue I see is that I believe we should change the structure of pythermalcomfort and have a separate file for each model. Which means that we can create a separate file for Hui and ATCS and there keep all the code related to those models, I would also refrain from having utils, coefficients, .... files for each model.

examples/calc_atcs.py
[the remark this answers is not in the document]
AkihisaNomoto: OK, added a description. I'll probably add another case of simulation later.

examples/calc_zhang_comfort_model.py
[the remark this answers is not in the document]
AkihisaNomoto: Added description and modified the example.

examples/jos3_output_example/jos3_example1 (default output).csv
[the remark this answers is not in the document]
AkihisaNomoto: Sorry I happed to run jos3 model. The version of pythermalcomfort has been updated, so "pythermalcomfort_version" column in this csv file has changed 2.8.4 from 2.8.2. Note that JOS3 model test code is all passing.

src/pythermalcomfort/zhang_comfort_model_functions/coefficients.py
AkihisaNomoto: Removed.

src/pythermalcomfort/zhang_comfort_model_functions/sensation.py
AkihisaNomoto: Reformatted.

src/pythermalcomfort/zhang_comfort_model_functions/sensation.py
AkihisaNomoto: Sorry about this. Added the description.

src/pythermalcomfort/zhang_comfort_model_functions/coefficients.py
AkihisaNomoto: I agree with you because Hui's model is not so complex even though I have to add more implementation. I put them together in a single file called "zhang_comfort.py".

src/pythermalcomfort/models.py
```diff
@@ -4633,3 +4643,676 @@ def bmr(self):
     def version(self):
         """version : float The current version of pythermalcomfort."""
         return self._version
+
+
+def zhang_sensation_comfort(
```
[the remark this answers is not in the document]
AkihisaNomoto: As we discussed today, I made new files for ATCS and Hui's model.

src/pythermalcomfort/zhang_comfort_model_functions/sensation.py
AkihisaNomoto: Added the comment. The default is "9-point".

src/pythermalcomfort/zhang_comfort_model_functions/coefficients_summary.csv
AkihisaNomoto: Oh, I see, that makes sense. I deleted the summary csv file.

src/pythermalcomfort/zhang_comfort_model_functions/utilities.py
AkihisaNomoto: I agree with you, but is it fine to do that later? 

My first priority for now is to get the latest models working on pythermalcomfort. I will make a new issue about this if you accept this initial pull request.

src/pythermalcomfort/zhang_comfort_model_functions/utilities.py
AkihisaNomoto: Thank you for pointing this out! Yes, there is a very similar function in JOS3 model. I will also make this cleaner later.

src/pythermalcomfort/zhang_comfort_model_functions/utilities.py
AkihisaNomoto: Moved to the main file.

src/pythermalcomfort/models.py
AkihisaNomoto: Thank you for your advise! That is the way!

src/pythermalcomfort/zhang_comfort_model_functions/coefficients.py
AkihisaNomoto: Agree that we should use dataclass, I will do it later.

AkihisaNomoto:
@FedericoTartarini Thanks for your great feedback. I moved Hui's model function and ATCS class to a separate file from models.py. I have not perfectly implemented what you advised, and I understand some parts of my code are a little terrible. However, since I have to show that the model is working on pythermalcomfort, could you please accept this pull request of this initial integration and add these model's info in `pythermalcomfort.rst`. Both model has docstrings, so this implementation should be straightforward. I hope you understand.

Again, thank you for all your support!

FedericoTartarini:
I may accept the pull request but in future, we should not rush it until we have fixed all the issues. People may start using the models once we publish a new version and then if we change the structure of the code, the code that other people have written will stop working and will break their implementation, this is very bad. Now many users are using pythermalcomfort hence we need to make sure the code is stable.
Can we keep the Pull on hold until we have fixed all the issues? You can mention in the presentation that we are planning to add the code to pythermalcomfort and in the meantime you can show the code in your fork.

The biggest issue at the moment is that now the ATCS and hui's model are within the pythermalcomfort folder and cannot imported within the model function which is very different from the other functions. Moreover, I am not sure that calling the model Zhan is the best way to go. Do people associate her name with a model? I am not sure about the latter. Probably the model should have a more meaningful name. We should also think carefully about the name ATCS. Once people start using it we cannot longer change that and ATCS to me means nothing. I would much prefer multi_node_comfort_model or advance_thermal_comfort_sensation, or something like that. 

What do you think? Sorry for being pedantic but we do not want to mess things up for other users.

FedericoTartarini:
@AkihisaNomoto I am now looking at how to create separate files for each of the models. i may have found a solution. Sorry to change the plan but could you please move back the ATCS and Hui's models to `models.py` so this is consistent with all the other functions, then I will work on moving them outside to their own file.

AkihisaNomoto:
I totally agree with you. Sorry, I was a bit in a rush yesterday because I will have rehearsal tomorrow. The way to import the model is different from the others so it would be a good idea to keep all the models in `models.py`. What do you think about the place where we put the coefficients for the equations? Any idea on this??

FedericoTartarini:
@AkihisaNomoto good news, I found a solution to the problem, I implemented it in the development branch. I added the heat_index() function to a folder called `functions` and created a separate test. I deployed this version to Test PyPI and it is working. 
```
pip install -i https://test.pypi.org/simple/ pythermalcomfort

from pythermalcomfort.functions import heat_index
print(heat_index(20,20))
```

Of course, this is not the final implementation since we want users to be able to import the function from models and not from functions so their old code will still work. However, this cannot be done until I remove the models.py file. If I have both a folder called `models` and a file `models.py` then I get a lot of errors so I will need to do that all in one go, where I move all the functions and I finally delete the models.py file.

I would, therefore, suggest that we keep the code as it is. We will not expose the functions you have added to the documentation so people will not start using them and then we can implement the above solution once we have a bit of time. I will, therefore, accept the pull request so it will be implemented when you present.

FedericoTartarini (approved):

FedericoTartarini:
@AkihisaNomoto I am merging the code into `master`. In the future, could you please work on the branch called `develpment` where we can edit the code and then we push to master only when we want to release a new version of the code.

PR #107, opened 2024-05-29
replace _set with set

ruijis:
The "set" is a more intuitive variable name, and it also correspond to the one In the documentation

### DIFF
--- a/pythermalcomfort/models/two_nodes.py
+++ b/pythermalcomfort/models/two_nodes.py
@@ -213,7 +213,7 @@ def two_nodes(
         "m_rsw": m_rsw,
         "w": w,
         "w_max": w_max,
-        "_set": _set,
+        "set": _set,
         "et": et,
         "pmv_gagge": pmv_gagge,
         "pmv_set": pmv_set,


FedericoTartarini:
I agree that we should consider changing this, however, we need to think carefully on how to implement this since this is a breaking change.

ruijis:
> I agree that we should consider changing this, however, we need to think carefully on how to implement this since this is a breaking change.

Can we update the document? Replacing the set with _set

FedericoTartarini:
@ruijis are you referring to update the documentation? I can do that but this will break old code, I will need to release a new major version of pythermalcomfort and explain the breaking change. I have some students currently working on the source code, maybe we can wait until they finish their work and then publish a new major version of pythermalcomfort which also includes this change. Do you need this to be done in a hurry?

ruijis:
> @ruijis are you referring to update the documentation? I can do that but this will break old code, I will need to release a new major version of pythermalcomfort and explain the breaking change. I have some students currently working on the source code, maybe we can wait until they finish their work and then publish a new major version of pythermalcomfort which also includes this change. Do you need this to be done in a hurry?

Never mind, I thought changing the documentation would be easy. 
If there's a new major version coming out, we should continue using set and modify the code.
Thank you, Fed.

FedericoTartarini:
We cannot only change the documentation since the output is called `_set` we will have to update both at the same time.

FedericoTartarini:
I have fixed this in version 3.0, I will publish this version soon

PR #120, opened 2024-09-13
Convert old unit test to url type

TwinGan:
@FedericoTartarini Professor, please review the code.


tests/test_a_pmv.py
```diff
@@ -1,10 +1,24 @@
...
+            inputs["tdb"],
+            inputs["tr"],
+            inputs["vr"],
+            inputs["rh"],
+            inputs["met"],
+            inputs["clo"],
+            inputs["a_coefficient"],
+        )
+        try:
+            assert is_equal(result, outputs["a_pmv"])
```
FedericoTartarini, line 19: should not be checking the tolerance here?

tests/test_ankle_draft.py
```diff
@@ -1,13 +1,23 @@
...
+    if test_data is None:
+        pytest.skip("Failed to retrieve test data")
 
-def test_ankle_draft():
-    assert (ankle_draft(25, 25, 0.2, 50, 1.2, 0.5, 0.3, units="SI")["PPD_ad"]) == 18.5
-    assert (
-        ankle_draft(77, 77, 0.2 * 3.28, 50, 1.2, 0.5, 0.4 * 3.28, units="IP")["PPD_ad"]
-    ) == 23.5
+    # Tests using data from URL
+    for case in test_data.get("data", []):
```
FedericoTartarini, line 11: the code should be the same across different test functions. in the test adaptive for example, you wrote the code

for entry in reference_table["data"]: while here you use the get function. I like this approach more `for entry in reference_table["data"]:` but we need to be consistent

tests/test_ankle_draft.py
```diff
@@ -1,13 +1,23 @@
...
+        inputs = case["inputs"]
+        print(case["outputs"])
```
FedericoTartarini, line 13: do not leave print in the code

tests/test_solar_gain.py
```diff
@@ -1,27 +1,18 @@
...
-                inputs["t_sol"],
-                inputs["f_svv"],
-                inputs["f_bes"],
-                inputs["asa"],
-                inputs["posture"],
-            )
-            assert sg["erf"] == outputs["erf"]
-            assert sg["delta_mrt"] == outputs["t_rsw"]
+def test_solar_gain(get_solar_gain_url, retrieve_data, is_equal):
+    reference_table = retrieve_data(get_solar_gain_url)
```
FedericoTartarini, line 5: there is a lot of code repetions, the function that extracts the:
* inputs
* outputs
* loops thought the keys and does the assertion 
are the same in all the test functions, we should create a function that basically does this automatically and we do not have to have code repetitions in all the tests

FedericoTartarini (commented):
@TwinGan I think the code is good but I have the following comments:
1. there are a lot of repetitions and we could create a function that does the test
2. sometimes you write slightly different code to achieve the same code, as if the code was written by two people. We should have consistent code. Please fix these and I will accept the pull request. Do not create some other test functions yet, let's keep working on the code you wrote here before moving ahead so we make sure the code is well written

tests/conftest.py
```diff
@@ -10,31 +10,73 @@
...
 @pytest.fixture
 def is_equal():
     def compare(a, b):
-        if (a is None and np.isnan(b)) or (b is None and np.isnan(a)):
+        if isinstance(a, np.ndarray):
+            if not isinstance(b, np.ndarray):
+                b = np.array(b, dtype=float)
+            b = np.where(b == None, np.nan, b)  # Replace None with np.nan
+            # Return True if arrays are close enough, including handling of NaN values
+            return np.allclose(a, b, equal_nan=True)
```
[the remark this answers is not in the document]
FedericoTartarini: we should pass atol = tolerance

tests/test_ankle_draft.py
```diff
@@ -1,13 +1,22 @@
...
+    if test_data is None:
+        pytest.skip("Failed to retrieve test data")
 
-def test_ankle_draft():
-    assert (ankle_draft(25, 25, 0.2, 50, 1.2, 0.5, 0.3, units="SI")["PPD_ad"]) == 18.5
-    assert (
-        ankle_draft(77, 77, 0.2 * 3.28, 50, 1.2, 0.5, 0.4 * 3.28, units="IP")["PPD_ad"]
-    ) == 23.5
+    # Tests using data from URL
+    for case in test_data.get("data", []):
```
FedericoTartarini, line 11: fix this

FedericoTartarini (approved):

FedericoTartarini (approved):

PR #123, opened 2024-09-23
Add tolerance for all test function

TwinGan:
1. Add tolerance for all test function
2. fix pmv_ppd test name error
all test cases passed
image

@FedericoTartarini

### DIFF
--- a/tests/test_clo_tout.py
+++ b/tests/test_clo_tout.py
@@ -1,43 +1,27 @@
-import unittest
 import numpy as np
 from pythermalcomfort.models import clo_tout
 
-
-class TestCloTout(unittest.TestCase):
-    def test_si_units_scalar(self):
-        # Test SI units with a scalar outdoor air temperature
-        result = clo_tout(tout=27)
-        self.assertEqual(result, 0.46)
-
-    def test_si_units_array(self):
-        # Test SI units with an array of outdoor air temperatures
-        result = clo_tout(tout=[27, 25])
-        np.testing.assert_equal(result, [0.46, 0.47])
-
-    def test_ip_units_scalar(self):
-        # Test IP units with a scalar outdoor air temperature
-        result = clo_tout(tout=80.6, units="IP")
-        self.assertEqual(result, 0.46)
-
-    def test_ip_units_array(self):
-        # Test IP units with an array of outdoor air temperatures
-        result = clo_tout(tout=[80.6, 77], units="IP")
-        np.testing.assert_equal(result, np.array([0.46, 0.47]))
-
-    def test_edge_cases(self):
-        # Test edge cases of the piecewise function
-        result = clo_tout(tout=[4, 2, 0])
-        np.testing.assert_equal(result, [0.67, 0.75, 0.82])
-
-        result = clo_tout(tout=[-4, -6])
-        np.testing.assert_equal(result, [0.96, 1.0])
-
-    def test_invalid_units(self):
-        # Test invalid units
-        with self.assertRaises(ValueError):
-            clo_tout(tout=27, units="invalid")
-
-    def test_invalid_type(self):
-        # Test invalid type for tout
-        with self.assertRaises(TypeError):
-            clo_tout(tout="invalid")
+def test_clo_tout(get_clo_tout_url, retrieve_data, is_equal):
+    
+    reference_table = retrieve_data(get_clo_tout_url)
+    
+    
+    for entry in reference_table["data"]:
+        inputs = entry["inputs"]
+        expected_output = entry["outputs"]["clo_tout"]
+        
+        
+        result = clo_tout(
+            tout=inputs["tout"],
+            units=inputs.get("units", "SI")  
+        )
+        
+        
+        try:
+            if isinstance(expected_output, list):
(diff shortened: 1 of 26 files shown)


tests/test_pet_steady.py
```diff
@@ -19,10 +18,7 @@ def test_pet_steady(get_pet_steady_url, retrieve_data, is_equal):
         )
 
         try:
-            if isinstance(expected_output, list):
-                np.testing.assert_equal(result, expected_output)
-            else:
-                assert np.isclose(result, expected_output, atol=0.1)
+            is_equal(result, expected_output, tolerance.get("PET", 1e-6))
```
FedericoTartarini, line 21: should not we use the `tolerance.get(key, ...` like in the other tests?

FedericoTartarini (approved):

FedericoTartarini:
@TwinGan great work, I left one minor comment and I will accept your pull request after I see your answer

tests/test_cooling_effect.py
```diff
@@ -1,32 +1,30 @@
...
         )
-        
+
         # To determine whether the result is as expected, use np.allclose for arrays and np.isclose for single
         try:
-            if isinstance(expected_output, list):
-                assert np.allclose(result, expected_output, atol=0.1), f"Expected {expected_output}, got {result}"
-            else:
-                assert np.isclose(result, expected_output, atol=0.1), f"Expected {expected_output}, got {result}"
+            is_equal(result, expected_output, tolerance.get("cooling_effect"))
```
FedericoTartarini, line 25: this should be fixed since if the key is not there then this may throw an issue.

tests/test_e_pmv.py
```diff
@@ -20,10 +20,7 @@ def test_e_pmv(get_e_pmv_url, retrieve_data, is_equal):
         )
 
         try:
-            if isinstance(expected_outputs, list):
-                np.testing.assert_equal(result, expected_outputs)
-            else:
-                assert is_equal(result, expected_outputs)
+            assert is_equal(result, expected_outputs, tolerance.get("pmv", 1e-6))
```
FedericoTartarini, line 23: please ensure consistency because sometimes you write the `key`name and sometimes you pass the key name as a string. I think the option with the key is better

tests/test_clo_tout.py
```diff
@@ -17,11 +17,12 @@ def test_clo_tout(get_clo_tout_url, retrieve_data, is_equal):
...
-        try:
-            if isinstance(expected_output, list):
-                np.testing.assert_equal(result, expected_output)
-            else:
-                assert is_equal(result, expected_output)
-        except AssertionError as e:
-            print(f"Assertion failed for clo_tout. Expected {expected_output}, got {result}, inputs={inputs}\nError: {str(e)}")
-            raise
\ No newline at end of file
+        for key in outputs:
```
FedericoTartarini, line 20: for key in tolerance??

tests/test_clo_tout.py
```diff
@@ -17,11 +17,12 @@ def test_clo_tout(get_clo_tout_url, retrieve_data, is_equal):
...
+            # Use the custom is_equal for other types
+            try:
+                assert is_equal(result, outputs[key], tolerance.get(key, 1e-6))
```
FedericoTartarini, line 23: I do not really like the get, we can simply use tolerance[key]

FedericoTartarini (commented):
I am submitting another review since you cannot see the comments

tests/test_cooling_effect.py
```diff
@@ -22,7 +22,7 @@ def test_cooling_effect(get_cooling_effect_url, retrieve_data, is_equal):
 
         # To determine whether the result is as expected, use np.allclose for arrays and np.isclose for single
         try:
-            is_equal(result, expected_output, tolerance.get("cooling_effect"))
+            is_equal(result, expected_output, tolerance.get("cooling_effect", 1e-6))
```
FedericoTartarini, line 25: we discussed this yesterday and we should remove the string cooling effect from here. Instead, we should be looping through the keys in tolerance and from there we should be getting the information about the tolerance. You have implemented this in some of the test functions

FedericoTartarini:
@TwinGan could you please reply comment and let me know if you have fixed all the issue, so I can then accept the pull request?

FedericoTartarini:
@TwinGan thank you so much for working on this. I have used all the changes you made as an inpiration to further simplify the code for the tests. Which now looks, for example:

```
from pythermalcomfort.models import a_pmv
from tests.conftest import Urls, retrieve_reference_table, validate_result


def test_a_pmv(get_test_url, retrieve_data):
    reference_table = retrieve_reference_table(
        get_test_url, retrieve_data, Urls.A_PMV.name
    )
    tolerance = reference_table["tolerance"]

    for entry in reference_table["data"]:
        inputs = entry["inputs"]
        outputs = entry["outputs"]
        result = a_pmv(**inputs)

        validate_result(result, outputs, tolerance)
```

as you can see the code is more simple and avoids many repetitions. I have pushed it to the `development` branch so you can have a look and review it. Please let me know what do you think and if you find any issues. I am going to close this pull request now since it is conflicting with the code I wrote.

PR #160, opened 2024-12-30
fix bug where t_mrt was incorrect for natural convection

jkittner:
the coefficient of exchange at the level of the globe was calculated incorrectly

This is taken from the ISO-norm: E DIN EN ISO 7726:2023-08 ISO/DIS 7726:2023(E)

grafik

This caused calculations of $T_{mrt}$ via the ISO-standard at low wind speed to be incorrect.

btw. I just saw that the function got moved and renamed - does that mean the new version will be completely incompatible with the previous versions? Without any deprecations?

### DIFF
--- a/pythermalcomfort/utilities.py
+++ b/pythermalcomfort/utilities.py
@@ -315,12 +315,12 @@ def mean_radiant_tmp(
 
         return np.around(tr, 1)
 
-    if standard == "iso":
+    if standard == "iso":  # pragma: no branch
         tg = np.add(tg, c_to_k)
         tdb = np.add(tdb, c_to_k)
 
         # calculate heat transfer coefficient
-        h_n = np.power(1.4 * (np.abs(tg - tdb) / d), 0.25)  # natural convection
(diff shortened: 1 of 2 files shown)


FedericoTartarini:
# coefficient of exchange at the level of the globe

Thank you for spotting this error, I will review the change and accept the pull request.

# new version

I am planning to release v3.0 soon which will include several breaking changes. I know this may be an issue for some users but I think this change is needed since I want to improve how the functions return the results and fix some other issues, like for example, the mean radiant temperature equation was erroneously added to the psychrometric module. I have described most of the breaking changes in the CHANGELOG to help users, however, please feel free to share tips on how I can help users with the migration to the new version.

pythermalcomfort/utilities.py
```diff
@@ -315,12 +315,12 @@ def mean_radiant_tmp(
 
         return np.around(tr, 1)
 
-    if standard == "iso":
+    if standard == "iso":  # pragma: no branch
```
FedericoTartarini, line 318: could you please help me understand the comment you have added?

Issue #176, opened 2025-02-17
Work capacity


FedericoTartarini:
Include the association between the work capacity and WBGT as a function of work intensity.

<img width="1565" alt="Image" src="https://github.com/user-attachments/assets/8e74d49b-aa79-4807-a52b-c26245abf58b" />

<img width="845" alt="Image" src="https://github.com/user-attachments/assets/47a1188a-12e7-486e-bd3b-40cc1a76490a" />

C-H-Simpson:
While you can use the functions from Todd Kjellstrom's work, I think there are functions with better empirical validity out there. These have made their way into a lot of studies (e.g. The Lancet Countdown, ILO reports) so I think it is useful to make them available. It is based on just two field studies and makes some slightly weird assumptions (in my opinion). "the empirical evidence is from studies in highly distinct locations, including a gold mine (Wyndham, 1969), 124 rice harvesters in West Bengal in India (Sahu et al., 2013), and six women observed in a climatic chamber (Nag and Nag, 1992)." (https://adaptecca.es/sites/default/files/documentos/2018_jrc_pesetaiii_impact_labour_productivity.pdf). The sigmoid form of the function is just an assumption, and the fit of the sigmoid to the data it is analysing is not especially good.

Foster, Smallcombe and others have a series of papers where they did some very well designed lab experiments to produce an empirical model of work capacity that can be applied to UTCI or WBGT. This has been used in a few other studies now.

1. Smallcombe JW, Foster J, Hodder SG, Jay O, Flouris AD, Havenith G. Quantifying the impact of heat on human physical work capacity; part IV: interactions between work duration and heat stress severity. Int J Biometeorol. 2022 Dec 1;66(12):2463–76. 
2. Foster J, Smallcombe JW, Hodder S, Jay O, Flouris AD, Nybo L, et al. Quantifying the impact of heat on human physical work capacity; part III: the impact of solar radiation varies with air temperature, humidity, and clothing coverage. Int J Biometeorol. 2022 Jan 1;66(1):175–88. 
3. Foster J, Smallcombe JW, Hodder S, Jay O, Flouris AD, Havenith G. Quantifying the impact of heat on human physical work capacity; part II: the observed interaction of air velocity with temperature, humidity, sweat rate, and clothing is not captured by most heat stress indices. International journal of biometeorology. 2022;66(3):507–20. 
[...]

PR #178, opened 2025-02-28
Add work capacity metrics

C-H-Simpson:
Aims to address issue #176 @FedericoTartarini 

I have added functions for estimating labour capacity loss. I have added a few references in the docstrings so people can trace the sources of the numbers. I have added some input validation, tests, and added it to the list of models.

Here is a plot of the functions I added:
workcapacity

It passed `tox -e py312` and `tox -e docs`.
You might want to check I implemented the input and output classes in a style consistent with the style of the repo as this is my first time contributing to this project.

### DIFF
--- a/docs/documentation/models.rst
+++ b/docs/documentation/models.rst
@@ -182,7 +182,7 @@ The Predicted Mean Vote (PMV) is an index that aims to predict the mean value of
 It was developed by Fanger [fanger1970]_.
 
 The PMV is designed to predict the average thermal sensation of a large group of people exposed to the same environment [ISO_7730_2005]_.
-It calculates the heat balance of a typical occupant and relates their thermal gains or losses to their predicted mean thermal sensation [ASHRAE_55_2023]_.
+It calculates the heat balance of a typical occupant and relates their thermal gains or capacityes to their predicted mean thermal sensation [ASHRAE_55_2023]_.
 
 The PMV can be used to check if a thermal environment meets comfort criteria and to establish requirements for different levels of acceptability.
 The PMV model is applicable to healthy men and women exposed to indoor environments where thermal comfort is desirable, but moderate deviations from thermal comfort occur, in the design of new environments or the assessment of existing ones.
@@ -325,3 +325,12 @@ Wind chill temperature
 .. autoclass:: pythermalcomfort.classes_return.WCT
     :members:
 
+Work capacity
+---------
+
+.. autofunction:: pythermalcomfort.models.work_capacity.workcapacity_dunne
+
+.. autofunction:: pythermalcomfort.models.work_capacity.workcapacity_hothaps
+
+.. autoclass:: pythermalcomfort.classes_return.WorkCapacity
+    :members:
--- a/docs/documentation/models.rst
+++ b/docs/documentation/models.rst
@@ -332,5 +332,8 @@ Work capacity
 
 .. autofunction:: pythermalcomfort.models.work_capacity.workcapacity_hothaps
 
+.. autofunction:: pythermalcomfort.models.work_capacity.workcapacity_iso
+
+.. autofunction:: pythermalcomfort.models.work_capacity.workcapacity_niosh
 .. autoclass:: pythermalcomfort.classes_return.WorkCapacity
     :members:
--- a/docs/documentation/models.rst
+++ b/docs/documentation/models.rst
@@ -182,7 +182,7 @@ The Predicted Mean Vote (PMV) is an index that aims to predict the mean value of
 It was developed by Fanger [fanger1970]_.
 
 The PMV is designed to predict the average thermal sensation of a large group of people exposed to the same environment [ISO_7730_2005]_.
-It calculates the heat balance of a typical occupant and relates their thermal gains or capacityes to their predicted mean thermal sensation [ASHRAE_55_2023]_.
+It calculates the heat balance of a typical occupant and relates their thermal gains or losses to their predicted mean thermal sensation [ASHRAE_55_2023]_.
 
 The PMV can be used to check if a thermal environment meets comfort criteria and to establish requirements for different levels of acceptability.
 The PMV model is applicable to healthy men and women exposed to indoor environments where thermal comfort is desirable, but moderate deviations from thermal comfort occur, in the design of new environments or the assessment of existing ones.
--- a/pythermalcomfort/classes_input.py
+++ b/pythermalcomfort/classes_input.py
@@ -52,6 +52,7 @@ class BaseInputs:
     max_skin_blood_flow: Union[float, int, np.ndarray, list] = field(default=80)
     max_sweating: Union[float, int, np.ndarray, list] = field(default=500)
     w_max: Union[float, int, np.ndarray, list] = field(default=None)
+    wbgt: Union[float, int, np.ndarray, list] = field(default=None)
 
     def __post_init__(self):
         def is_pandas_series(obj):
@@ -216,6 +217,9 @@ class BaseInputs:
         if self.w_max is not None:
             self.w_max = convert_series_to_list(self.w_max)
             validate_type(self.w_max, "w_max", (float, int, np.ndarray, list))
+        if self.wbgt is not None:
+            self.wbgt = convert_series_to_list(self.wbgt)
+            validate_type(self.wbgt, "wbgt", (float, int, np.ndarray, list))
 
 
 @dataclass
@@ -850,4 +854,15 @@ class WCTInputs(BaseInputs):
             tdb=tdb,
             v=v,
             round_output=round_output,
+        )
+
+@dataclass
+class WorkCapacityInputs(BaseInputs):
+    def __init__(
+        self,
+        wbgt,
+    ):
+        # Initialize with only required fields, setting others to None
+        super().__init__(
+            wbgt=wbgt,
         )
--- a/pythermalcomfort/classes_input.py
+++ b/pythermalcomfort/classes_input.py
@@ -53,6 +53,7 @@ class BaseInputs:
     max_sweating: Union[float, int, np.ndarray, list] = field(default=500)
     w_max: Union[float, int, np.ndarray, list] = field(default=None)
     wbgt: Union[float, int, np.ndarray, list] = field(default=None)
+    intensity: Union[str, np.ndarray, list] = field(default=None)
 
     def __post_init__(self):
         def is_pandas_series(obj):
@@ -856,13 +857,30 @@ class WCTInputs(BaseInputs):
             round_output=round_output,
         )
 
+
 @dataclass
-class WorkCapacityInputs(BaseInputs):
+class WorkCapacityHothapsInputs(BaseInputs):
     def __init__(
         self,
         wbgt,
+        intensity,
     ):
         # Initialize with only required fields, setting others to None
         super().__init__(
             wbgt=wbgt,
+            intensity=intensity,
+        )
+
+
+@dataclass
+class WorkCapacityStandardsInputs(BaseInputs):
+    def __init__(
+        self,
+        wbgt,
+        met,
+    ):
+        # Initialize with only required fields, setting others to None
+        super().__init__(
+            wbgt=wbgt,
+            met=met,
         )
--- a/pythermalcomfort/classes_return.py
+++ b/pythermalcomfort/classes_return.py
@@ -638,6 +638,22 @@ class WCT:
 
 
 @dataclass(frozen=True)
+class WorkCapacity:
+    """Dataclass to represent work loss.
+
+    Attributes
+    ----------
+    capacity : float or list of floats
+        Work capacity affected by heat.
+    """
+
+    capacity: Union[float, npt.ArrayLike]
+
+    def __getitem__(self, item):
+        return getattr(self, item)
+
+
+@dataclass(frozen=True)
 class JOS3BodyParts:
     """Dataclass to represent the body parts in the JOS3 model. It is very important to
     keep the order of the attributes as they are defined in the dataclass ['head',
--- a/pythermalcomfort/models/work_capacity.py
+++ b/pythermalcomfort/models/work_capacity.py
@@ -2,12 +2,14 @@ from typing import Union
 
 import numpy as np
 
+from pythermalcomfort.classes_input import WorkCapacityInputs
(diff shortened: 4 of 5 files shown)


docs/documentation/models.rst
```diff
@@ -325,3 +325,15 @@ Wind chill temperature
 .. autoclass:: pythermalcomfort.classes_return.WCT
     :members:
 
+Work capacity
```
FedericoTartarini, line 328: Could you please change this heading to 
```
Work Capacity
===========
```

Below please briefly describe what these model are trying to predict.

docs/documentation/models.rst
```diff
@@ -325,3 +325,15 @@ Wind chill temperature
...
+---------
+
+.. autofunction:: pythermalcomfort.models.work_capacity.workcapacity_dunne
```
FedericoTartarini, line 331: Please add name of the model, for example

Dunne
------

docs/documentation/models.rst
```diff
@@ -325,3 +325,15 @@ Wind chill temperature
...
+
+.. autofunction:: pythermalcomfort.models.work_capacity.workcapacity_hothaps
```
FedericoTartarini, line 333: see my previous comment

docs/documentation/models.rst
```diff
@@ -325,3 +325,15 @@ Wind chill temperature
...
+
+.. autofunction:: pythermalcomfort.models.work_capacity.workcapacity_iso
+
+.. autofunction:: pythermalcomfort.models.work_capacity.workcapacity_niosh
+.. autoclass:: pythermalcomfort.classes_return.WorkCapacity
```
FedericoTartarini, line 338: maybe move this at the top if it is shared across all models, right under the description of the work capacity models

pythermalcomfort/classes_input.py
```diff
@@ -216,6 +218,9 @@ def convert_series_to_list(obj):
         if self.w_max is not None:
             self.w_max = convert_series_to_list(self.w_max)
             validate_type(self.w_max, "w_max", (float, int, np.ndarray, list))
+        if self.wbgt is not None:
```
FedericoTartarini, line 221: please also implement the post_init for intensity

pythermalcomfort/classes_return.py
```diff
@@ -637,6 +637,22 @@ def __getitem__(self, item):
...
+class WorkCapacity:
+    """Dataclass to represent work loss.
+
+    Attributes
+    ----------
+    capacity : float or list of floats
+        Work capacity affected by heat.
+    """
+
+    capacity: Union[float, npt.ArrayLike]
```
FedericoTartarini, line 650: in the other classess I am using `Union[float, list[float]]` but I guess what you implemented is better

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+import numpy as np
+
+from pythermalcomfort.classes_input import (
+    WorkCapacityHothapsInputs,
+    WorkCapacityStandardsInputs,
+)
+from pythermalcomfort.classes_return import WorkCapacity
+
+
+def workcapacity_dunne(wbgt: Union[float, list[float]], intensity: str) -> WorkCapacity:
```
FedericoTartarini, line 12: please split these functions into separate files, like I am doing for all the other models. I guess you can simply call the model `dunne` no need to add `workcapacity`

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+    """
+    Estimate work capacity due to heat based ISO standards as described by Dunne et al
```
FedericoTartarini, line 14: please add reference, example `Returns the value of the Cooling Effect (`CE`_) calculated in compliance
    with the ASHRAE 55 2020 Standard [ASHRAE_55_2023]_.`

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Heavy intensity work is sometimes labelled as 400 W, but this is only
+    nominal. Medium work is assumed to be half as much as heavy intensity, and
+    light half as much as medium.
+
+    Parameters
+    ----------
+    wbgt : float or list of floats
+        Wet bulb globe temperature, [°C].
+    intensity : str
```
FedericoTartarini, line 32: could you please better explain quantitatively what this adjectives mean?

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+        `capacity` value, use the `capacity` attribute of the returned `WorkCapacity` instance, e.g.,
+        `result.capacity`.
+
+    """
+    # Validate inputs
+    WorkCapacityHothapsInputs(wbgt=wbgt, intensity=intensity)
+
+    wbgt = np.array(wbgt)
+    capacity = np.clip((100 - (25 * (np.maximum(0, wbgt - 25)) ** (2 / 3))), 0, 100)
+    if intensity == "heavy":
```
FedericoTartarini, line 50: I guess you can remove this if statement since it does not do anything

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+        raise ValueError("intensity should = heavy, med, or light")
+
+    return WorkCapacity(capacity=capacity)
+
+
+def workcapacity_hothaps(
+    wbgt: Union[float, list[float]], intensity: str
+) -> WorkCapacity:
+    """
+    Estimate work capacity due to heat based on Kjellstrom et al.
```
FedericoTartarini, line 66: please add citation

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+    nominal. Medium work is assumed to be half as much as heavy intensity, and
+    light half as much as medium.
+
+    Parameters
+    ----------
+    wbgt : float or list of floats
+        Wet bulb globe temperature, [°C].
+    intensity : str
+        Which work intensity to use for the calculation, choice of "heavy",
+        "med" or "light".
```
FedericoTartarini, line 34: should `med` be medium for consistency?

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+
+    """
+    # Validate inputs
+    WorkCapacityHothapsInputs(wbgt=wbgt, intensity=intensity)
+
+    wbgt = np.array(wbgt)
+    capacity = np.clip((100 - (25 * (np.maximum(0, wbgt - 25)) ** (2 / 3))), 0, 100)
+    if intensity == "heavy":
+        pass
+    elif intensity == "med":
```
FedericoTartarini, line 52: we should convert these strings to a class, especially beacause many models use the same inputs. For exmaple, in the `classess_input` file you can add
```
class WorkIntensities(Enum):
    low = "low"
    moderate = "moderate"
heavy = 'heavy'
```
and then you can use this dataclass in these if statements and in the other funcitons below

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Note that for this function "the empirical evidence is from studies in
+    heavyly distinct locations, including a gold mine (Wyndham, 1969), 124 rice
+    harvesters in West Bengal in India (Sahu et al., 2013), and six women
+    observed in a climatic chamber (Nag and Nag, 1992)."
+    (https://adaptecca.es/sites/default/files/documentos/2018_jrc_pesetaiii_impact_labour_productivity.pdf).
+    The shape of the function is just an assumption, and the fit of the
+    sigmoid to the data it is analysing is not especially good.
+
+    Heavy intensity work is sometimes labelled as 400 W, medium 300 W, light 200
```
FedericoTartarini, line 82: W/m2 or W?

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+    Estimates the amount of work that will be done at a given WBGT and
+    intensity of work as a percent. 100% means work is unaffected by heat. 0%
+    means no work is done.
+
+    This is based upon NIOSH safety standards. See:
+    Dunne JP, Stouffer RJ, John JG. Reductions in labour capacity from heat
+    stress under climate warming. Nature Climate Change. 2013 Jun;3(6):563–6.
+
+    Heavy intensity work is sometimes labelled as 400 W, but this is only
+    nominal. Medium work is assumed to be half as much as heavy intensity, and
```
FedericoTartarini, line 25: could you please provide the values med=200W. Could you please also explain what `nominal` means in this context?

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+        `result.capacity`.
+
+    """
+    # Validate inputs
+    WorkCapacityHothapsInputs(wbgt=wbgt, intensity=intensity)
+
+    wbgt = np.array(wbgt)
+    capacity = np.clip((100 - (25 * (np.maximum(0, wbgt - 25)) ** (2 / 3))), 0, 100)
+    if intensity == "heavy":
+        pass
```
FedericoTartarini, line 51: in the function inputs please set heavy as default value since this is what the code does.

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+    elif intensity == "med":
+        capacity = np.clip(capacity * 2, 0, 100)
+    elif intensity == "light":
+        capacity = np.clip(capacity * 4, 0, 100)
```
FedericoTartarini, line 55: as you did in the function below, you can do the clip at the end. It makes the code easier to read

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+    For a comparison of different functions see Fig 1 of Day E, Fankhauser S, Kingsmill
+    N, Costa H, Mavrogianni A. Upholding labour productivity under climate
+    change: an assessment of adaptation options. Climate Policy. 2019
+    Mar;19(3):367–85.
+
+    Parameters
+    ----------
+    wbgt : float or list of floats
+        Wet bulb globe temperature, [°C].
+    M : float or list of floats
```
FedericoTartarini, line 162: 1. inputs should be lower case to comply with PEP. 
2. Are you sure it is in W not in W/m2 like for all the other models? For consistency I am wondring if we should ask for an input in `met` since all the other models require the input to be in met and not W

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+
+    """
+    # Validate inputs
+    WorkCapacityStandardsInputs(wbgt=wbgt, met=M)
+
+    wbgt = np.array(wbgt)
+    M = np.array(M)
+    if (M < 0).any() or (M > 2500).any():
+        raise ValueError("Metabolic rate out of plausible range")
+    Mrest = 117  # assumed resting metabolic rate
```
FedericoTartarini, line 182: please call it met_rest

pythermalcomfort/models/work_capacity.py
```diff
@@ -0,0 +1,243 @@
...
+    For a comparison of different functions see Fig 1 of Day E, Fankhauser S, Kingsmill
+    N, Costa H, Mavrogianni A. Upholding labour productivity under climate
+    change: an assessment of adaptation options. Climate Policy. 2019
+    Mar;19(3):367–85.
+
+    Parameters
+    ----------
+    wbgt : float or list of floats
+        Wet bulb globe temperature, [°C].
+    M : float or list of floats
```
FedericoTartarini, line 216: see previous comment

tests/test_workcapacity.py
FedericoTartarini, line 1: 1. we should split the tests into separate files, one for each model
2. more importantly than testing for value errors we should actually tests that the model gives the right output or different inputs, like for example assert that workcapacity_dunne(30, "heavy") == XX, but also test for medium and low
3. we should also test when the user pass an array as input to make sure the functions work

FedericoTartarini (changes requested):
Amazing work, thank you so much for contributing and helping writing the code. I really appreciate your time and help.

I left a few comments to be addressed before I can accept the pull request. If some are not clear, please reply in line with the comment. I can also help you edit the code, if needed.

pythermalcomfort/models/work_capacity.py
C-H-Simpson: Brode et al write "metabolic heat production (M) given in Watts." https://link.springer.com/article/10.1007/s00484-017-1346-9

I think in the ISO standard, it is actually in terms of W/m2 (see doi.org/10.2486/indhealth.44.368, I don't have access to the actual standard). The numbers in Table 1 therein do not match the numbers that Brode used, but they are close-ish if you assume a body area of about 2 m2.

So, I think Brode's numbers are in W not W/m2, despite the ISO standard being in W/m2. I'm a little dissatisfied that Brode's numbers don't match though. I will try to make it clearer that this is Brode et al's interpretation of the standard and it might not match the tables in the standard.

C-H-Simpson: As for using MET units rather than Watts... This will mean we will need to take an input of body area and mass too, which seems like over-complicating it. It might make more sense to implement a utility function that estimates metabolic rate in Watts from MET units.

I agree it seems potentially confusing if all other functions have an input "met" that is in MET, but that this function has one called "met" in Watts.

FedericoTartarini: > I agree it seems potentially confusing if all other functions have an input "met" that is in MET, but that this function has one called "met" in Watts.

I am worried about this since some of the users will not carefully read the documentation and then this will cause an erorr.

> As for using MET units rather than Watts... This will mean we will need to take an input of body area and mass too, which seems like over-complicating it.

While it is true that we are going to make the function a bit more complicated, I believe that most of the users will only know the MET value and not the total power (W). Consequently, they will have to calculate the W anyway. Consequently, we can allow the user to pass only MET and we use the default weight and height to calculate W. If the user wants, they can also modify weight and height by passing them as an input.

FedericoTartarini:
Closing this one since @rcqz has opened a new one #195

PR #184, opened 2025-04-22
Update phs

rcqz:
# Description

This PR includes updates to the PHS model in compliance with the ISO 7933:2023 standard. 

## Main Changes in the New Standard

The main changes include: 
- `sw_max` no longer depends on `met`
- `hc_dyn` is calculated with `t_cl` instead of `t_sk`
- default values changed for `f_r`, `t_re` and `t_cr_eq`
- `d_lim_loss` and `d_max` uses a single value instead of 50 and 95 quantiles
- formula for `f_cl_r` changed sligntly

## Testing
Test cases have been implemented according to Annex F from the ISO 7933:2023 standard. At the moment all tests except for the third one pass with the following tolerance:
- water loss: 2.5%
- `t_cr`: 0.2
- `d_lim_loss`: 10

## Summary by CodeRabbit

- **New Features**
  - Added support for the updated ISO 7933:2023 standard to the Predicted Heat Strain (PHS) model, allowing users to select between the 2004 and 2023 versions.
  - Updated documentation and references to include ISO 7933:2023.

- **Bug Fixes**
  - Corrected calculations and parameter handling in the PHS model for improved accuracy and compliance with the latest standard.

- **Tests**
  - Added comprehensive tests to validate the PHS model's behavior under the 2023 standard.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,11 +1,12 @@
 Changelog
 =========
 
-3.1.0 ()
+3.1.0 (2025-04-23)
 -------------------
 * Updated the PHS model in compliance with the ISO 7933:2023 standard
     - Added default‑kwarg overrides for 2023 mode (f_r, t_re, t_cr_eq)
     - removed unused variable `round` from `default_kwargs`
+* Included test cases according to the ISO 7933:2023 standard
 
 3.0.1 (2025-04-14)
 -------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+3.1.0 ()
+-------------------
+* Updated the PHS model in compliance with the ISO 7933:2023 standard
+    - Added default‑kwarg overrides for 2023 mode (f_r, t_re, t_cr_eq)
+    - removed unused variable `round` from `default_kwargs`
+
 3.0.1 (2025-04-14)
 -------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,10 @@ Changelog
 * Updated the PHS model in compliance with the ISO 7933:2023 standard
     - Added default‑kwarg overrides for 2023 mode (f_r, t_re, t_cr_eq)
     - removed unused variable `round` from `default_kwargs`
+    - By default the ISO 7933:2023 standard will now be used. The 2004 standard can still be specified as an optional parameter
 * Included test cases according to the ISO 7933:2023 standard
+* WARNING: The third ISO 7933:2023 test case is currently failing and has been marked xfail while the discrepancy is investigated.
+
 
 3.0.1 (2025-04-14)
 -------------------
--- a/pythermalcomfort/models/phs.py
+++ b/pythermalcomfort/models/phs.py
@@ -97,17 +97,18 @@ def phs(
     duration : int, optional
         Duration of the work sequence, [minutes]. Defaults to 480.
     f_r : float, optional
-        Emissivity of the reflective clothing, [dimensionless]. Defaults to 0.97 in the 2004 standard and 0.42 in the 2023 standard.
+        Emissivity of the reflective clothing, [dimensionless]. Defaults to 0.97 in the 2004 standard and
+        0.42 in the 2023 standard.
     t_sk : float, optional
         Mean skin temperature when worker starts working, [°C]. Defaults to 34.1.
     t_cr : float, optional
         Mean core temperature when worker starts working, [°C]. Defaults to 36.8.
     t_re : float, optional
-        Mean rectal temperature when worker starts working, [°C]. If False in the 2004 standard, then t_re = t_cr, whereas in the 2023 standard t_re = 36.8 °C
+        Mean rectal temperature when worker starts working, [°C]. If False in the 2004 standard,
+        then t_re = t_cr, whereas in the 2023 standard t_re = 36.8 °C
     t_cr_eq : float, optional
-        Mean core temperature as a function of met when worker starts working, [°C]. If
-        False in the 2004 standard, then t_cr_eq = t_cr, whereas in the 2023 standard
-        t_cr_eq = 36.8 °C.
+        Mean core temperature as a function of met when worker starts working, [°C]. If False in the 2004
+        standard, then t_cr_eq = t_cr, whereas in the 2023 standard t_cr_eq = 36.8 °C.
     sweat_rate : float, optional
         Initial sweat rate, [g/h]. Defaults to 0.
 
@@ -143,7 +144,8 @@ def phs(
 
     if model not in [Models.iso_7933_2004.value, Models.iso_7933_2023.value]:
         raise ValueError(
-            f"Model should be '{Models.iso_7933_2004.value}' or '{Models.iso_7933_2023.value}'. Please check the documentation."
+            f"Model should be '{Models.iso_7933_2004.value}' or '{Models.iso_7933_2023.value}'. "
+            "Please check the documentation."
         )
 
     PHSInputs(
@@ -218,7 +220,7 @@ def phs(
 
     if model == Models.iso_7933_2023.value:
         p_a = 0.6105 * np.exp(17.27 * tdb / (tdb + 237.3)) * rh / 100
-    elif model == Models.iso_7933_2004.value:
+    else:  # model == Models.iso_7933_2004.value:
         p_a = p_sat(tdb) / 1000 * rh / 100
 
     acclimatized = int(acclimatized)
@@ -369,10 +371,10 @@ def _phs_optimized(
     d_lim_loss_95 = 0
 
     if model == Models.iso_7933_2023.value:
-        # 2023 standard only has one Dmax value
+        # 2023 standard only has one d_max value
         d_max_50 = (0.03 if drink == 0 else 0.05) * weight * 1000
         d_max_95 = (0.03 if drink == 0 else 0.05) * weight * 1000
-    elif model == Models.iso_7933_2004.value:
+    else:  # model == Models.iso_7933_2004.value:
         # maximum water loss to protect a mean subject [g]
         d_max_50 = 0.075 * weight * 1000
         # maximum water loss to protect 95 % of the working population [g]
@@ -394,7 +396,7 @@ def _phs_optimized(
         a_r_du = 0.77
     elif posture == Postures.sitting.value:
         a_r_du = 0.7
-    elif posture == Postures.crouching.value:
+    else:  # posture == Postures.crouching.value:
         a_r_du = 0.67
 
     # evaluation of the max sweat rate as a function of the metabolic rate
@@ -406,7 +408,7 @@ def _phs_optimized(
             sw_max = 250
         if acclimatized == 100:
             sw_max = sw_max * 1.25
-    elif model == Models.iso_7933_2023.value:
+    else:  # model == Models.iso_7933_2023.value:
         sw_max = 400 if acclimatized == 0 else 500
 
     # max skin wettedness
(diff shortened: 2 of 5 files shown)


FedericoTartarini:
@coderabbitai review

CHANGELOG.rst
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+3.1.0 (2025-04-23)
+-------------------
+* Updated the PHS model in compliance with the ISO 7933:2023 standard
+    - Added default‑kwarg overrides for 2023 mode (f_r, t_re, t_cr_eq)
+    - removed unused variable `round` from `default_kwargs`
+* Included test cases according to the ISO 7933:2023 standard
```
FedericoTartarini, line 9: please add a bullet point saying that test case 3 from the standard is not passing.

pythermalcomfort/models/phs.py
```diff
@@ -59,6 +60,8 @@ def phs(
         external work, [met] default 0
     round_output : bool, optional
         If True, rounds output value. If False, it does not round it. Defaults to True.
+    model : str, optional
+        Select the model you want to use to calculate the PHS. The default option is "7933-2004", and the other option is "7933-2023".
```
FedericoTartarini, line 64: change this accordingly. You will also have to change the tests

pythermalcomfort/models/phs.py
```diff
@@ -88,19 +91,19 @@ def phs(
...
     f_r : float, optional
-        Emissivity of the reflective clothing, [dimensionless]. Defaults to 0.97.
+        Emissivity of the reflective clothing, [dimensionless]. Defaults to 0.97 in the 2004 standard and 0.42 in the 2023 standard.
     t_sk : float, optional
         Mean skin temperature when worker starts working, [°C]. Defaults to 34.1.
     t_cr : float, optional
         Mean core temperature when worker starts working, [°C]. Defaults to 36.8.
     t_re : float, optional
-        Mean rectal temperature when worker starts working, [°C]. Defaults to False.
+        Mean rectal temperature when worker starts working, [°C]. Defaults to False in the 2004 standard and 36.8 in the 2023 standard.
```
FedericoTartarini, line 104: this and the following change about t_cr_eq are not 100 correct since even in the old standard this had a default value, since the code was as follows. 

```
    if not t_re:
        t_re = t_cr
    if not t_cr_eq:
        t_cr_eq = t_cr
```

hence we can change the docs to `Mean rectal temperature when worker starts working, [°C]. If False, in the 2004 standard then t_re = t_cr while in the 2023 standard t_re = 36.8 °C`

pythermalcomfort/models/phs.py
```diff
@@ -88,19 +91,19 @@ def phs(
...
     t_cr_eq : float, optional
-        Mean core temperature as a function of met when worker starts working, [°C]. Defaults to False.
+        Mean core temperature as a function of met when worker starts working, [°C]. Defaults to False in the 2004 standard and 36.8 in the 2023 standard.
```
FedericoTartarini, line 106: change this to reflect the comment above.

pythermalcomfort/models/phs.py
```diff
@@ -133,6 +136,12 @@ def phs(
         )
         print(result.t_re)  # [37.5 43.4]
     """
+
+    if model not in [Models.iso_7933_2004.value, Models.iso_7933_2023.value]:
+        raise ValueError(
+            "Model should be '7933-2004' or '7933-2023'. Please check the documentation."
```
FedericoTartarini, line 142: change to f"Model should be '{Models.iso_7933_2004.value}' or '{Models.iso_7933_2023.value}'. Please check the documentation."

pythermalcomfort/models/phs.py
```diff
@@ -153,7 +162,7 @@ def phs(
         "height": 1.8,
         "walk_sp": 0,
         "theta": 0,
-        "acclimatized": 100,
+        "acclimatized": 1,
```
FedericoTartarini, line 165: keep 100

pythermalcomfort/models/phs.py
```diff
@@ -194,11 +212,14 @@ def phs(
     sweat_rate = kwargs["sweat_rate"]
     limit_inputs = kwargs["limit_inputs"]
 
-    p_a = p_sat(tdb) / 1000 * rh / 100
+    if model == Models.iso_7933_2023.value:
+        p_a = 0.6105 * math.exp(17.27 * tdb / (tdb + 237.3)) * rh / 100
```
FedericoTartarini, line 216: please check if this value is in fact different from p_a = p_sat(tdb) / 1000 * rh / 100. If they are very close no need to add a new equation.

pythermalcomfort/models/phs.py
```diff
@@ -194,11 +212,14 @@ def phs(
...
+    elif model == Models.iso_7933_2004.value:
+        p_a = p_sat(tdb) / 1000 * rh / 100
 
     acclimatized = int(acclimatized)
-    if acclimatized < 0 or acclimatized > 100:
-        raise ValueError("Acclimatized should be between 0 and 100")
+    if acclimatized not in [0, 1]:
```
FedericoTartarini, line 221: revert

pythermalcomfort/models/phs.py
```diff
@@ -208,9 +229,6 @@ def phs(
             "The weight of the person should be in kg and it cannot exceed 1000"
         )
 
-    # I needed to create this variable since vectorize accept no more than 34 inputs+outputs
-    combined_drink_acc_weight = drink * 1000000 + acclimatized * 1000 + weight
```
FedericoTartarini, line 212: why have you removed this? Is the code now working? Before it was not since as stated in the comment `I needed to create this variable since vectorize accept no more than 34 inputs+outputs`

pythermalcomfort/models/phs.py
```diff
@@ -266,6 +287,11 @@ def phs(
     }
 
     if limit_inputs:
+        if model == Models.iso_7933_2004.value:
```
FedericoTartarini, line 290: this code is not needed, we do not need to introduce a new variable called standard, see comment below.

pythermalcomfort/models/phs.py
```diff
@@ -274,7 +300,7 @@ def phs(
             met_valid,
             clo_valid,
         ) = _check_standard_compliance_array(
-            "7933", tdb=tdb, tr=tr, v=v, met=met, clo=clo, p_a=p_a
+            standard, tdb=tdb, tr=tr, v=v, met=met, clo=clo, p_a=p_a
```
FedericoTartarini, line 303: here simply pass model as the first input

pythermalcomfort/models/phs.py
```diff
@@ -357,30 +391,35 @@ def _phs_optimized(
         def_speed = 1
 
     # radiating area dubois
-    a_r_du = 0.7
     if posture == Postures.standing.value:
         a_r_du = 0.77
-    if posture == Postures.crouching.value:
+    elif posture == Postures.sitting.value:
```
FedericoTartarini, line 396: This change will introduce a warning in PyCharm since the variable may not be defined. Perhaps the previous code was better.

pythermalcomfort/models/phs.py
```diff
@@ -357,30 +391,35 @@ def _phs_optimized(
...
-        sw_max = 250
-    if acclimatized >= 50:
-        sw_max = sw_max * 1.25
+    if model == Models.iso_7933_2004.value:
+        sw_max = (met - 32) * a_dubois
+        if sw_max > 400:
+            sw_max = 400
+        if sw_max < 250:
+            sw_max = 250
+        if acclimatized == 1:
```
FedericoTartarini, line 408: change to 100

pythermalcomfort/models/phs.py
```diff
@@ -357,30 +391,35 @@ def _phs_optimized(
...
+            sw_max = sw_max * 1.25
+    elif model == Models.iso_7933_2023.value:
+        sw_max = 400 if acclimatized == 0 else 500
 
     # max skin wettedness
-    if acclimatized < 50:
-        w_max = 0.85
-    else:
-        w_max = 1
+    w_max = 0.85 if acclimatized == 0 else 1
```
FedericoTartarini, line 414: change to 100

pythermalcomfort/models/phs.py
```diff
@@ -510,9 +558,13 @@ def _phs_optimized(
         elif w_req >= 1.7:
             sw_req = sw_max
         else:
-            e_v_eff = 1 - w_req**2 / 2
             if w_req > 1:
                 e_v_eff = (2 - w_req) ** 2 / 2
+            else:
+                e_v_eff = 1 - w_req**2 / 2
```
FedericoTartarini, line 564: this is very strange since the equation for w_req > 1 is completely different from the equation in the else. In one we do 2 - w and then we do at the power of 2 in the other we have w**2. Could you please double check this?

FedericoTartarini: Moreover, this equation was not in the 2004 version of the standard, are you sure that this will not impact the results of the 2004 model?

pythermalcomfort/utilities.py
```diff
@@ -450,7 +452,7 @@ def _check_standard_compliance_array(standard, **kwargs):
 
         return values_to_return.values()
 
-    if standard == "7933":  # based on ISO 7933:2004 Annex A
+    if standard == "7933_2004":  # based on ISO 7933:2004 Annex A
```
FedericoTartarini, line 455: change this to if standard == Models.iso_7933_2004.value

pythermalcomfort/utilities.py
```diff
@@ -460,6 +462,16 @@ def _check_standard_compliance_array(standard, **kwargs):
 
         return tdb_valid, tr_valid, v_valid, p_a_valid, met_valid, clo_valid
 
+    if standard == "7933_2023":  # based on ISO 7933:2023 Annex A
```
FedericoTartarini, line 465: same as previous comment

tests/test_phs.py
```diff
@@ -18,6 +19,145 @@ def test_phs(get_test_url, retrieve_data):
...
+    result = phs(**inputs)
+
+    # check water_loss and core temperature
+    assert result.water_loss == pytest.approx(expected["water_loss"], rel=0.025)
+    assert result.t_cr == pytest.approx(expected["t_cr"], abs=0.3)
+
+    assert result.d_lim_loss_95 == pytest.approx(expected["d_lim_loss_95"], abs=10)
+
+    if "d_lim_t_re" in expected:
+        assert result.d_lim_t_re == pytest.approx(expected["d_lim_t_re"], abs=10)
```
FedericoTartarini, line 158: See code rabbit comment and please Temporarily mark test 3 as expected to fail until a fix can be implemented

FedericoTartarini (commented):
@rcqz great work and thank you for working on this. I have left a few comments, could you please review them one by one and mark them as completed if resolved or write a comment if what I am suggesting is not clear?

tests/test_phs.py
```diff
@@ -18,6 +19,145 @@ def test_phs(get_test_url, retrieve_data):
...
+            ),
+            dict(
+                water_loss=6538,
+                t_cr=37.6,
+                d_lim_loss_95=280,
+            ),
+        ),
+        (
+            dict(
+                tdb=35,
```
rcqz, line 54: Also add an array as input test

pythermalcomfort/models/phs.py
rcqz: They give very close values for temperatures >= 0 Celsius (about 0.1% difference). But for negative values they start to differ. For example at -10°C there's a 10% relative difference and at -20°C there's a 20% difference.

pythermalcomfort/models/phs.py
rcqz: I agree this is confusing as the equation are not consistant in the 2023 standard itself - between the pseudocode they provided and the equations they defined in the document. Here I implemented it according to the pseudocode they have provided.

pythermalcomfort/models/phs.py
rcqz: I tried to include these variables separately and they pass all the test cases at the moment.

FedericoTartarini (approved):

PR #185, opened 2025-04-25
Auto str mixin

rcqz:
A simple base class is added so that the model outputs are formatted.

## Summary by CodeRabbit

- **New Features**
  - Updated the Predicted Heat Strain (PHS) model to support the ISO 7933:2023 standard, with new default parameters and the ability to specify the 2004 standard if needed.
  - Introduced a new, consistently formatted display for results from all dataclasses, making outputs easier to read.

- **Tests**
  - Added new test cases for the updated PHS model based on the ISO 7933:2023 standard, with one test case currently marked as expected to fail.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+3.1.0 (2025-04-23)
+-------------------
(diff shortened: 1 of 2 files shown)


FedericoTartarini:
@coderabbitai review

FedericoTartarini (approved):

PR #303, opened 2026-04-14
Fix/sports heat stress risk scale

yehui-h:
1. fix: bump version to 3.9.2 and update sports heat stress risk scale from 0.0-3.0 to 1.0-4.0
2. fix: enforce minimum sport air speed in sports_heat_stress_risk

## Release Notes – Version 3.9.2

* **New Features**
  * Updated sports heat stress risk scale from 0.0–3.0 to 1.0–4.0 with revised risk thresholds and recommendations.
  * Calculations now apply sport-specific minimum air speed thresholds.

* **Chores**
  * Version bumped to 3.9.2.

### DIFF
--- a/pythermalcomfort/models/sports_heat_stress_risk.py
+++ b/pythermalcomfort/models/sports_heat_stress_risk.py
@@ -112,12 +112,14 @@ def sports_heat_stress_risk(
     rh : float or list of float
         Relative humidity [%].
     vr : float or list of float
-        Relative air speed [m/s].
+        Relative air speed [m/s]. Values lower than ``sport.vr`` are set to
+        ``sport.vr`` before the risk calculation.
     sport : _SportsValues
         Sport-specific activity dataclass with fields ``clo`` (clothing insulation),
-        ``met`` (metabolic rate), ``vr`` (relative air speed), and ``duration`` (activity duration).
-        Use one of the predefined entries from the :py:class:`Sports` class, e.g., ``Sports.RUNNING``,
-        ``Sports.SOCCER``, ``Sports.TENNIS``, etc.
+        ``met`` (metabolic rate), ``vr`` (minimum relative air speed),
+        and ``duration`` (activity duration). Use one of the predefined entries from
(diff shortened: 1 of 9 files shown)


pythermalcomfort/models/sports_heat_stress_risk.py
```diff
@@ -112,12 +112,14 @@ def sports_heat_stress_risk(
     rh : float or list of float
         Relative humidity [%].
     vr : float or list of float
-        Relative air speed [m/s].
+        Relative air speed [m/s]. Values lower than ``sport.vr`` are set to
```
FedericoTartarini: please change this to: "Relative air speed [m/s]. Please note that if the input vr is lower than the self generated wind speed for each sport defined in XX, then the sport.vr will be used for the calculation."

FedericoTartarini (changes requested):
I'm not entirely sure what is the best way of updating the version moving forward whether it is better that I update the version once I push it to main or whether in each pull request we should update the version number let me know what you think is the best approach.

Perhaps if we update it in the Pull request to development then I'm not entirely sure it is going to be deployed when we push it to main. I need to check the GitHub action.

yehui-h:
> I'm not entirely sure what is the best way of updating the version moving forward whether it is better that I update the version once I push it to main or whether in each pull request we should update the version number let me know what you think is the best approach.
> 
> Perhaps if we update it in the Pull request to development then I'm not entirely sure it is going to be deployed when we push it to main. I need to check the GitHub action.

1. For developers using this in other projects, the most important thing is whether PyPI already has a released version that includes the change, and whether their dependency constraints allow that version to be installed.

2. If they are not installing from PyPI but directly from a GitHub tag or branch, then the corresponding tag, commit, or branch version becomes what matters most.

3. Currently, this project is automatically published to PyPI via GitHub Actions "build-test-publish.yml" when a commit is pushed to the master branch.

yehui-h:
Once it has been merged into development, it can then be merged into master and deployed, so that I can use it properly in SMA app.

FedericoTartarini (approved):

PR #307, opened 2026-04-28
fix: add units to printed UTCI output

### LINKED ISSUE #295: Add units to print class ouput
Current behaviour below. We should be adding the variable either as an additional column or just next to the variable name. For example `utci  [C]          : [24.60, 40.60]`

```python
print(result)

------------------------------------------------------------------
                               UTCI                               
------------------------------------------------------------------
utci            : [24.60, 40.60]
stress_category : ['no thermal stress', 'very strong heat stress']
```

Tong-Yuzhou:
## Summary

This PR addresses #295 by adding unit labels to printed result outputs when field units are defined.

Currently, UTCI printed output shows the field name without a unit. This PR adds a lightweight `_field_units` mapping to `AutoStrMixin`, so result classes can define units for specific printed fields.

For UTCI, the printed output now includes:

```text
utci [°C]
```

while non-numeric category fields such as `stress_category` remain unchanged.

## Changes

- Added `_field_units` support to `AutoStrMixin`.
- Updated printed field names to include units when defined.
- Added the UTCI unit mapping for the `utci` field.
- Added a test to verify that `utci [C]` appears in the printed output.

## Tests

I ran the following checks locally:

```bash
pipenv run python -m pytest tests/test_classes_return.py -q
pipenv run python -m pytest -k utci
pipenv run ruff check --fix pythermalcomfort tests
pipenv run ruff format pythermalcomfort tests
pipenv run python -m pytest -q
```

Results:

```text
300 passed, 1 xfailed, 7 warnings
```

Closes #295.

## Summary by CodeRabbit

* **New Features**
  * Result objects now automatically display unit annotations for applicable fields in their string representations, enhancing clarity when viewing thermal comfort outputs. Temperature and other relevant values now include unit labels (such as "[C]" for Celsius) to provide better context at a glance.

* **Tests**
  * Added test coverage verifying unit annotation formatting displays correctly in result outputs.

### DIFF
--- a/pythermalcomfort/classes_return.py
+++ b/pythermalcomfort/classes_return.py
@@ -681,7 +681,7 @@ class UTCI(AutoStrMixin):
         UTCI categorized in terms of thermal stress [Blazejczyk2013]_.
     """
 
-    _field_units: ClassVar[dict[str, str]] = {"utci": "C"}
+    _field_units: ClassVar[dict[str, str]] = {"utci": "°C"}
 
     utci: float | list[float]
     stress_category: str | list[str]
--- a/pythermalcomfort/classes_return.py
+++ b/pythermalcomfort/classes_return.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import datetime as dt
 import textwrap
 from dataclasses import dataclass, fields, is_dataclass
+from typing import ClassVar
 
 import numpy as np
 import numpy.typing as npt
@@ -10,6 +11,8 @@ from numpy._typing import NDArray
 
 
 class AutoStrMixin:
+    _field_units: ClassVar[dict[str, str]] = {}
+
     def __str__(self) -> str:
         """Pretty-print dataclass fields as aligned single-line 'name : value'.
 
@@ -23,6 +26,10 @@ class AutoStrMixin:
         MAX_STR_WIDTH = 80
         names = [f.name for f in fields(type(self))]
         values = [getattr(self, n) for n in names]
+        display_names = [
+            f"{name} [{self._field_units[name]}]" if name in self._field_units else name
+            for name in names
+        ]
 
         # Format values as strings, using comma separation for arrays/lists
         def value_to_str(val):
@@ -52,11 +59,11 @@ class AutoStrMixin:
 
         value_strs = [value_to_str(v) for v in values]
         # Find max name length for alignment
-        max_name_len = max(len(n) for n in names) if names else 0
+        max_name_len = max(len(n) for n in display_names) if display_names else 0
         # Compose lines with aligned colons
         lines = [
             f"{n.ljust(max_name_len)} : {v}"
-            for n, v in zip(names, value_strs, strict=True)
+            for n, v in zip(display_names, value_strs, strict=True)
         ]
         # Shorten each line to MAX_STR_WIDTH
         lines = [
@@ -673,6 +680,8 @@ class UTCI(AutoStrMixin):
     stress_category : str or list of strs
         UTCI categorized in terms of thermal stress [Blazejczyk2013]_.
     """
+
+    _field_units: ClassVar[dict[str, str]] = {"utci": "C"}
 
     utci: float | list[float]
     stress_category: str | list[str]
--- a/tests/test_classes_return.py
+++ b/tests/test_classes_return.py
@@ -85,6 +85,13 @@ def test_autostr_includes_field_units() -> None:
 
     output = str(result)
 
-    assert "utci [C]" in output
+    assert "utci [°C]" in output
     assert "stress_category" in output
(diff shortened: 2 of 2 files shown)


pythermalcomfort/classes_return.py
```diff
@@ -674,6 +681,8 @@ class UTCI(AutoStrMixin):
         UTCI categorized in terms of thermal stress [Blazejczyk2013]_.
     """
 
+    _field_units: ClassVar[dict[str, str]] = {"utci": "°C"}
```
FedericoTartarini, line 684: Thank you very much for implementing this, however I'm not a huge fan of using dictionaries because they are very difficult to maintain because they rely on strings Another issue that I see here is that we are defining the unit in the classes return and not in the model Python function. I'm wondering if this is the best approach and if this is going to be easy to implement. What is going to happen if a model accepts different inputs in SI and IP units and if we only define the unit here in the classes return?

FedericoTartarini (changes requested):
Thank you very much for working on this, I really appreciate your contribution. However, I would like to explore different opportunities and ways on how to implement this. This is mainly because we don't need to implement this only for the UTCI function but for all the other models. Consequently, I'm not entirely sure whether defining the units in a dictionary in the classes return in the auto string mixing is the best approach. I haven't thought about how to implement these but I want to carefully think about it before implementing it. Could you please explore a few options on how we can implement this? As previously noted, please mention that if implemented this, we need to implement it for all the models with all the different outputs that they may have. Perhaps the best approach is to define a unit in the classes return for each individual model but again, I need to think about this and I have not had the chance to find an elegant solution to this problem

pythermalcomfort/classes_return.py
Tong-Yuzhou: Thank you very much for the detailed feedback. I agree that the current implementation is too narrow and that a string-keyed `_field_units` dictionary is probably not the best long-term solution.

I see a few possible approaches:

1. Field-level unit metadata in the return dataclasses

Each output field could declare its unit directly in the dataclass field definition, for example through a small helper such as `unit_field(...)`.

This would keep the unit information close to the output field and avoid a separate dictionary based on string field names. `AutoStrMixin` would only read the field metadata and format the printed output.

This approach seems cleaner than the current dictionary-based implementation, but we would still need a way to handle SI/IP-dependent units.

2. Pass display units from each model function to the return object

Each model function already knows the selected `units` argument, such as SI or IP. Therefore, the model function could pass the appropriate display units to the return object.

This would make the unit decision more explicit and closer to the computation logic. It may be the most accurate approach for models that support both SI and IP units, but it would require changes across many model functions.

3. Define a centralized unit schema or registry

We could define standard unit categories, such as temperature, velocity, pressure, percentage, heat flux, and metabolic rate. Return fields would then be linked to these unit categories instead of raw strings.

This would be more systematic and easier to keep consistent across all models, but it may be a larger design change.

4. First prepare an inventory of all return classes and output fields

Before changing the implementation, I could first prepare a list of all return classes, their output fields, and their expected SI/IP units. This would make it easier to evaluate which implementation is most suitable for the whole package, not only UTCI.

My current preference would be either option 1 or option 2, or a combination of both: field-level metadata for each output variable, with the selected SI/IP unit system passed from the model function when needed.

I would be happy to revise the PR based on the approach you think is most appropriate.

FedericoTartarini: I think that the most important point that we should reflect is that currently we are already defining the units of the output in the docstring. Consequently, we want to be able to show the units when we print the results, but we also need to make sure that the units in the doc string and in the outputs remain in sync and always updated now and in the future.

I think that starting from point 4 would allow us to do an informed decision and understand which approach is best. I like approach 1 because it's probably the cleanest and the simplest. It would also be very nice to see a few code examples on how this can be implemented. We could potentially even have this code example here, or even better in the issue. So we can discuss it over there with the other collaborators and then decide which solution is the best and the most appropriate. We also need to have a look and understand how much the unit input actually affects the output.

Tong-Yuzhou: Thank you very much for the clarification. I agree that the current dictionary-based implementation is not the best long-term solution.
I will move the design discussion to issue #295 and prepare a small inventory of return classes/output fields, together with a few possible implementation examples, so that we can compare the options before changing the package-wide implementation.

FedericoTartarini: Thank you very much for looking into this. Shall we close this pull request? And then you can open a new one when ready.

Tong-Yuzhou: Yes, that sounds good. Thank you very much for the guidance.
I will close this pull request for now and continue the design discussion in issue #295. Once the approach is clearer and I have prepared the inventory and implementation examples, I will open a new PR.

PR #291, opened 2026-03-19
Feature/add ireq model

ChuanLi32781:
Add IREQ  model based on ISO 11079

## Summary by CodeRabbit

* **New Features**
  * Added ISO 11079 Required Clothing Insulation (IREQ) and Duration Limited Exposure (DLE) calculations and made them available via the public API.

* **Documentation**
  * Added a dedicated IREQ/DLE documentation section and corrected several punctuation/formatting and encoding issues across docs and changelog.

* **Examples**
  * Added an example script demonstrating scalar and array-based usage.

* **Tests**
  * Added tests covering scalar/array results, broadcasting, clamping, validation and public API exposure.

PR #315, opened 2026-05-10
fix: add UserWarning when inputs exceed model applicability limits

### LINKED ISSUE #297: value outside the model limits
Currently if a user passes a value which is outside the model applicability limits we don't return any warning we just return the value as "nan" perhaps we should be adding at least a warning in this case.

JitaoFu:
## What this PR does

- Add `param_name` argument to `valid_range()` so it emits a `UserWarning` with the parameter name, out-of-range value(s), valid range, and affected indices when `limit_inputs=True`
- Update `_check_standard_compliance_array()` to pass `param_name` for all standards, and add targeted warnings for the three `airspeed_control=False` conditions in ASHRAE 55
- Move compliance checks inside `if limit_inputs:` blocks in `adaptive_ashrae`, `adaptive_en`, `pmv_ppd_iso`, and `pmv_ppd_ashrae` so warnings are suppressed when `limit_inputs=False`

## Why this change is needed

Closes #297: users had no visibility into which inputs were silently clamped or ignored when values fell outside a model's applicability limits. The new warnings surface the parameter name, the offending value(s), the valid range, and the affected row indices, making debugging far easier.

## How I tested it

- Added `TestValidRange` unit tests covering `valid_range()` warning behaviour
- Added `TestCheckStandardComplianceArray` tests covering all three `airspeed_control=False` conditions in ASHRAE 55
- Added tests for `heat_index_rothfusz` and `pmv_ppd_ashrae` out-of-range warning paths
- Ran the full test suite

## Notes for reviewers

- For models exposing a limit_inputs parameter (e.g. pmv_ppd_ashrae, adaptive_ashrae, pmv_ppd_iso, utci), warnings are suppressed when limit_inputs=False. Helper functions without this parameter (ankle_draft, vertical_tmp_grad_ppd, mean_radiant_tmp with standard="Mixed Convection") always warn unconditionally.
- Warning message format:

```
  UserWarning: Some values of <param_name> are outside the applicability limits <limits> and will be set to NaN. Out-of-range values: <values> at indices <indices>.
```

### DIFF
--- a/pythermalcomfort/models/adaptive_ashrae.py
+++ b/pythermalcomfort/models/adaptive_ashrae.py
@@ -113,14 +113,6 @@ def adaptive_ashrae(
             v=v,
         )
 
-    tdb_valid, tr_valid, v_valid = _check_standard_compliance_array(
-        Models.ashrae_55_2023.value,
-        tdb=tdb,
-        tr=tr,
-        v=v,
-    )
-    trm_valid = valid_range(t_running_mean, (10.0, 33.5))
-
     to = operative_tmp(tdb, tr, v, standard=standard)
 
     # Calculate cooling effect (ce) of elevated air speed when top > 25 degC.
@@ -133,6 +125,13 @@ def adaptive_ashrae(
     t_cmf = 0.31 * t_running_mean + 17.8
 
     if limit_inputs:
+        tdb_valid, tr_valid, v_valid = _check_standard_compliance_array(
+            Models.ashrae_55_2023.value,
+            tdb=tdb,
+            tr=tr,
+            v=v,
+        )
+        trm_valid = valid_range(t_running_mean, (10.0, 33.5), "t_running_mean")
         all_valid = ~(
             np.isnan(tdb_valid)
             | np.isnan(tr_valid)
--- a/pythermalcomfort/models/adaptive_ashrae.py
+++ b/pythermalcomfort/models/adaptive_ashrae.py
@@ -6,9 +6,8 @@ from pythermalcomfort.classes_input import ASHRAEInputs
 from pythermalcomfort.classes_return import AdaptiveASHRAE
 from pythermalcomfort.shared_functions import valid_range
 from pythermalcomfort.utilities import (
-    Models,
     Units,
-    _check_standard_compliance_array,
+    _check_ashrae55_compliance,
     operative_tmp,
     units_converter,
 )
@@ -125,13 +124,12 @@ def adaptive_ashrae(
     t_cmf = 0.31 * t_running_mean + 17.8
 
     if limit_inputs:
-        tdb_valid, tr_valid, v_valid = _check_standard_compliance_array(
-            Models.ashrae_55_2023.value,
+        tdb_valid, tr_valid, v_valid = _check_ashrae55_compliance(
             tdb=tdb,
             tr=tr,
             v=v,
         )
-        trm_valid = valid_range(t_running_mean, (10.0, 33.5), "t_running_mean")
+        trm_valid = valid_range(t_running_mean, (10.0, 33.5))
         all_valid = ~(
             np.isnan(tdb_valid)
             | np.isnan(tr_valid)
--- a/pythermalcomfort/models/ankle_draft.py
+++ b/pythermalcomfort/models/ankle_draft.py
@@ -111,12 +111,16 @@ def ankle_draft(
     if units.upper() == Units.IP.value:
         tdb, tr, vr, v_ankle = units_converter(tdb=tdb, tr=tr, vr=vr, vel=v_ankle)
 
-    tdb_valid, tr_valid, v_valid, v_limited = _check_standard_compliance_array(
-        standard=Models.ashrae_55_2023.value,
-        tdb=tdb,
-        tr=tr,
-        v_limited=vr,
-        v=vr,
+    tdb_valid, tr_valid, v_valid, met_valid, clo_valid, v_limited = (
+        _check_standard_compliance_array(
+            standard=Models.ashrae_55_2023.value,
+            tdb=tdb,
+            tr=tr,
+            v_limited=vr,
+            v=vr,
+            met=met,
+            clo=clo,
+        )
     )
 
     if np.all(np.isnan(v_limited)):
@@ -132,6 +136,7 @@ def ankle_draft(
         met,
         clo,
         model=Models.ashrae_55_2023.value,
+        limit_inputs=False,
     ).pmv
     ppd_val = np.around(
         np.exp(-2.58 + 3.05 * v_ankle - 1.06 * tsv)
@@ -140,4 +145,15 @@ def ankle_draft(
         1,
     )
     acceptability = ppd_val <= 20
+
+    all_valid = ~(
+        np.isnan(tdb_valid)
+        | np.isnan(tr_valid)
+        | np.isnan(v_valid)
+        | np.isnan(met_valid)
+        | np.isnan(clo_valid)
+    )
+    ppd_val = np.where(all_valid, ppd_val, np.nan)
+    acceptability = np.where(all_valid, acceptability, np.nan)
+
     return AnkleDraft(ppd_ad=ppd_val, acceptability=acceptability)
--- a/pythermalcomfort/models/ankle_draft.py
+++ b/pythermalcomfort/models/ankle_draft.py
@@ -22,6 +22,7 @@ def ankle_draft(
     clo: float | list[float],
     v_ankle: float | list[float],
     units: str = Units.SI.value,
+    limit_inputs: bool = True,
 ) -> AnkleDraft:
     """Calculate the percentage of thermally dissatisfied people with the ankle draft
     (0.1 m) above floor level [Liu2017]_.
@@ -71,6 +72,12 @@ def ankle_draft(
 
     units : {'SI', 'IP'}
         Select the SI (International System of Units) or the IP (Imperial Units) system.
+    limit_inputs : bool, optional
+        By default, if the inputs are outside the standard applicability limits the
+        function returns nan. If False, returns values even if input values are
+        outside the applicability limits of the model. Defaults to True. The
+        applicability limits are 10 < tdb [°C] < 40, 10 < tr [°C] < 40,
+        0 < vr [m/s] < 0.2, 1 < met [met] < 4, and 0 < clo [clo] < 1.5.
 
     Returns
     -------
@@ -97,6 +104,7 @@ def ankle_draft(
         clo=clo,
         v_ankle=v_ankle,
         units=units,
+        limit_inputs=limit_inputs,
     )
 
     # Convert lists to numpy arrays
@@ -110,23 +118,6 @@ def ankle_draft(
 
     if units.upper() == Units.IP.value:
         tdb, tr, vr, v_ankle = units_converter(tdb=tdb, tr=tr, vr=vr, vel=v_ankle)
-
-    tdb_valid, tr_valid, v_valid, met_valid, clo_valid, v_limited = (
-        _check_ashrae55_compliance(
-            tdb=tdb,
-            tr=tr,
-            v_limited=vr,
-            v=vr,
-            met=met,
-            clo=clo,
-            v_param_name="vr",
-        )
-    )
-
-    if np.all(np.isnan(v_limited)):
-        raise ValueError(
-            "This equation is only applicable for air speed lower than 0.2 m/s",
-        )
 
     tsv = pmv_ppd_ashrae(
         tdb,
@@ -146,14 +137,32 @@ def ankle_draft(
     )
     acceptability = ppd_val <= 20
 
-    all_valid = ~(
-        np.isnan(tdb_valid)
-        | np.isnan(tr_valid)
-        | np.isnan(v_valid)
-        | np.isnan(met_valid)
-        | np.isnan(clo_valid)
-    )
-    ppd_val = np.where(all_valid, ppd_val, np.nan)
-    acceptability = np.where(all_valid, acceptability, np.nan)
+    if limit_inputs:
+        tdb_valid, tr_valid, v_valid, met_valid, clo_valid, v_limited = (
+            _check_ashrae55_compliance(
+                tdb=tdb,
+                tr=tr,
+                v_limited=vr,
+                v=vr,
+                met=met,
+                clo=clo,
+                v_param_name="vr",
+            )
+        )
+
+        if np.all(np.isnan(v_limited)):
+            raise ValueError(
+                "This equation is only applicable for air speed lower than 0.2 m/s",
+            )
+
+        all_valid = ~(
+            np.isnan(tdb_valid)
+            | np.isnan(tr_valid)
+            | np.isnan(v_valid)
+            | np.isnan(met_valid)
+            | np.isnan(clo_valid)
+        )
+        ppd_val = np.where(all_valid, ppd_val, np.nan)
+        acceptability = np.where(all_valid, acceptability, np.nan)
 
     return AnkleDraft(ppd_ad=ppd_val, acceptability=acceptability)
--- a/pythermalcomfort/models/ankle_draft.py
+++ b/pythermalcomfort/models/ankle_draft.py
@@ -8,7 +8,7 @@ from pythermalcomfort.models.pmv_ppd_ashrae import pmv_ppd_ashrae
 from pythermalcomfort.utilities import (
     Models,
     Units,
-    _check_standard_compliance_array,
+    _check_ashrae55_compliance,
     units_converter,
 )
 
@@ -112,14 +112,14 @@ def ankle_draft(
         tdb, tr, vr, v_ankle = units_converter(tdb=tdb, tr=tr, vr=vr, vel=v_ankle)
 
     tdb_valid, tr_valid, v_valid, met_valid, clo_valid, v_limited = (
-        _check_standard_compliance_array(
-            standard=Models.ashrae_55_2023.value,
+        _check_ashrae55_compliance(
             tdb=tdb,
             tr=tr,
             v_limited=vr,
             v=vr,
             met=met,
             clo=clo,
+            v_param_name="vr",
         )
     )
 
--- a/pythermalcomfort/models/pmv_ppd_iso.py
+++ b/pythermalcomfort/models/pmv_ppd_iso.py
@@ -9,7 +9,6 @@ from pythermalcomfort.shared_functions import mapping, valid_range
 from pythermalcomfort.utilities import (
     Models,
     Units,
-    _check_standard_compliance_array,
     units_converter,
 )
 
@@ -162,29 +161,21 @@ def pmv_ppd_iso(
             "PMV calculations can only be performed in compliance with ISO 7730-2005",
         )
 
-    pmv_array = _pmv_ppd_optimized(tdb, tr, vr, rh, met, clo, wme)
+    pmv = _pmv_ppd_optimized(tdb, tr, vr, rh, met, clo, wme)
 
     ppd_array = 100.0 - 95.0 * np.exp(
-        -0.03353 * pmv_array**4.0 - 0.2179 * pmv_array**2.0,
+        -0.03353 * pmv**4.0 - 0.2179 * pmv**2.0,
     )
 
     # Checks that inputs are within the bounds accepted by the model if not return nan
     if limit_inputs:
-        (
-            tdb_valid,
-            tr_valid,
-            v_valid,
-            met_valid,
-            clo_valid,
-        ) = _check_standard_compliance_array(
-            standard=Models.iso_7730_2005.value,
-            tdb=tdb,
-            tr=tr,
-            v=vr,
-            met=met,
-            clo=clo,
-        )
-        pmv_valid = valid_range(pmv_array, (-2, 2))  # this is the ISO limit
+        # ISO 7730-2005 page 3 applicability limits
+        tdb_valid = valid_range(tdb, (10.0, 30.0))
+        tr_valid = valid_range(tr, (10.0, 40.0))
+        v_valid = valid_range(vr, (0.0, 1.0))
+        met_valid = valid_range(met, (0.8, 4.0))
+        clo_valid = valid_range(clo, (0.0, 2.0))
+        pmv_valid = valid_range(pmv, (-2, 2))
 
         all_valid = ~(
             np.isnan(tdb_valid)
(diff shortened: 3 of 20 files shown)


TIAN-TOM:
@JitaoFu could you retarget this one to `development` when you get a chance? It looks like the PR workflow and CodeRabbit are only wired up for PRs targeting `development`, so nothing is running while the base is `master`. Federico asked for the same base change on #291 on 8 May 2026.

I checked the branches locally, and `development` and `master` have the same tree right now, so the PR diff should not change.

One small thing in the PR description too: it says warnings only emit when `limit_inputs=True`. That is true for the models you gated, but a few helpers without that flag now warn as well, including `ankle_draft`, `vertical_tmp_grad_ppd`, and `mean_radiant_tmp(..., standard="Mixed Convection")`. Could you adjust that sentence so it does not overstate the behaviour?

TIAN-TOM:
@JitaoFu I noticed one user-facing issue while checking this locally.

`ankle_draft` and `vertical_tmp_grad_ppd` already call `_check_standard_compliance_array`, then call `pmv_ppd_ashrae(...)`, which runs the same check again. With an out-of-range `tdb`, users see the same warning twice. Could you pass `limit_inputs=False` to the inner `pmv_ppd_ashrae` call, since the outer check has already handled it?

A smaller cleanup: in `pmv_ppd_ashrae.py`, the `tdb_valid`, `tr_valid`, `v_valid`, `met_valid`, and `clo_valid` variables are created in one `if limit_inputs:` block and used in a later one. It works, but keeping the compliance handling in one block would make the function easier to maintain.

JitaoFu:
@TIAN-TOM ,Thanks for the review. Here's what has been addressed:

**Duplicate warnings in `ankle_draft` and `vertical_tmp_grad_ppd`**: Fixed. 
- For `ankle_draft`, the outer compliance check was expanded to include `met` and `clo`, an `all_valid` output mask was added, and `limit_inputs=False` was passed to the inner `pmv_ppd_ashrae` call. 
- For `vertical_tmp_grad_ppd`, only `limit_inputs=False` was needed — the existing `v_limited` (0–0.2) check is already stricter than the inner `v` (0–2.0) check, so the outer coverage was already complete.

**Split `if limit_inputs:` blocks in `pmv_ppd_ashrae`**: Refactored. 
- `tdb_ce`/`tr_ce`/`vr_ce` are now used for CE-corrected values, keeping the original variables intact. This allows the compliance check and output masking to sit in a single `if limit_inputs:` block after computation, consistent with `adaptive_ashrae`, `pmv_ppd_iso`, and `utci`.                  

**PR description**
- Updated to distinguish between functions with a `limit_inputs` parameter (warnings gated by the flag) and helper functions without one (`ankle_draft`, `vertical_tmp_grad_ppd`, `mean_radiant_tmp` with `standard="Mixed Convection"`) that always warn unconditionally.

pythermalcomfort/models/adaptive_ashrae.py
```diff
@@ -133,6 +125,13 @@ def adaptive_ashrae(
     t_cmf = 0.31 * t_running_mean + 17.8
 
     if limit_inputs:
+        tdb_valid, tr_valid, v_valid = _check_standard_compliance_array(
+            Models.ashrae_55_2023.value,
+            tdb=tdb,
+            tr=tr,
+            v=v,
+        )
+        trm_valid = valid_range(t_running_mean, (10.0, 33.5), "t_running_mean")
```
FedericoTartarini: The only thing that I don't like of this approach is that inside valid range we need to pass the array of values, the range and then the variable names as a string which is kind of a repetition because that's the same as the name of the variable in the first input.It would be ideal if you could kind of get the signature and the name of the variable and then just print that. So valid range will accept the variable and the range as before, but then we print the variable name if the output is outside the limit.Does this make sense? I don't know if this can be done.

pythermalcomfort/models/ankle_draft.py
FedericoTartarini, line 74: Isn't this function missing the limit input input?

pythermalcomfort/models/pmv_ppd_iso.py
```diff
@@ -185,6 +170,20 @@ def pmv_ppd_iso(
 
     # Checks that inputs are within the bounds accepted by the model if not return nan
     if limit_inputs:
+        (
+            tdb_valid,
+            tr_valid,
+            v_valid,
+            met_valid,
+            clo_valid,
+        ) = _check_standard_compliance_array(
```
FedericoTartarini: Shall we consider deleting the function called checkStandardComplianceArray and use the default valid range to  be more  consistent?

pythermalcomfort/models/pmv_ppd_iso.py
```diff
@@ -185,6 +170,20 @@ def pmv_ppd_iso(
...
+            standard=Models.iso_7730_2005.value,
+            tdb=tdb,
+            tr=tr,
+            v=vr,
+            met=met,
+            clo=clo,
+        )
         pmv_valid = valid_range(pmv_array, (-2, 2))  # this is the ISO limit
```
FedericoTartarini: I guess that this is missing the third input "pmv"

pythermalcomfort/models/vertical_tmp_grad_ppd.py
FedericoTartarini, line 67: This function is missing the limit inputs.

FedericoTartarini: I think that this code can be moved inside the if statement to check if the image inputs are true or false.

pythermalcomfort/shared_functions.py
```diff
@@ -1,12 +1,36 @@
+import warnings
 from collections.abc import Mapping
 from typing import Any
 
 import numpy as np
 
 
-def valid_range(x, valid) -> np.ndarray:
+def valid_range(x, valid, param_name=None) -> np.ndarray:
```
FedericoTartarini: See my previous comment, I really like this solution however it would be fantastic if instead of passing the parameter name we could extract this somehow from the name of the variable x. Could you please look into this solution and see if this is even possible?

pythermalcomfort/shared_functions.py
```diff
@@ -1,12 +1,36 @@
...
     """Filter values based on a valid range."""
-    return np.where((x >= valid[0]) & (x <= valid[1]), x, np.nan)
+    x_arr = np.asarray(x)
+    out_of_range_mask = (x_arr < valid[0]) | (x_arr > valid[1])
+
+    if param_name is not None and np.any(out_of_range_mask):
+        if x_arr.ndim == 0:  # scalar input
+            warnings.warn(
+                f"Value of '{param_name}' ({float(x_arr.item())}) is outside the "
```
FedericoTartarini: Please install pre-commits on your computer, there is instructions on how to do that in the documentation. I think that you are going to get an error with this code here, because the message of the warning should be defined as a variable first and then passed to the warnings. For example you can say msg = f"... and then pass this message to the warning.

pythermalcomfort/shared_functions.py
```diff
@@ -1,12 +1,36 @@
...
+                f"applicability limits [{valid[0]}, {valid[1]}] and will be "
+                f"set to NaN.",
+                UserWarning,
+                stacklevel=2,
+            )
+        else:  # array input
+            bad_indices = np.flatnonzero(out_of_range_mask).tolist()
+            bad_values = x_arr[out_of_range_mask].tolist()
+            warnings.warn(
+                f"Some values of '{param_name}' are outside the applicability "
```
FedericoTartarini: This message could be reformatted because we know how many bad values we have so instead of saying some values we should say x where x is the number of variables of parameter name outside applicability limits. Also I've seen that there are a lot of repetitions in these strings Can we combine for example line 26 with line 16 or do we need to define them separately?

pythermalcomfort/shared_functions.py
```diff
@@ -1,12 +1,36 @@
...
+                f"Value of '{param_name}' ({float(x_arr.item())}) is outside the "
+                f"applicability limits [{valid[0]}, {valid[1]}] and will be "
+                f"set to NaN.",
+                UserWarning,
+                stacklevel=2,
+            )
+        else:  # array input
+            bad_indices = np.flatnonzero(out_of_range_mask).tolist()
+            bad_values = x_arr[out_of_range_mask].tolist()
+            warnings.warn(
```
FedericoTartarini: Please see my previous and comment after this one but to avoid repetitions inside the if and else we should just define the message and then pass it to the warnings warn outside the if statement because at the end of the day the warnings.warn is the same for both of the cases.

pythermalcomfort/utilities.py
FedericoTartarini: Please check my previous comment but I'm actually wondering whether we should delete this function and only keep the valid range function. I think this function is unnecessarily complete and also it does not check all the models. So I'm just wondering whether we need it at all. Please check it carefully and don't delete it just because you wrote this comment but critically assess whether it is better to delete it or keep it.

pythermalcomfort/utilities.py
```diff
@@ -458,93 +456,117 @@ def _check_standard_compliance_array(standard, **kwargs):
...
 
         if not params["airspeed_control"]:
-            v_valid = np.where(
-                (params["v"] > 0.8) & (params["clo"] < 0.7) & (params["met"] < 1.3),
-                np.nan,
-                v_valid,
-            )
+            cond1 = (params["v"] > 0.8) & (params["clo"] < 0.7) & (params["met"] < 1.3)
+            if np.any(cond1):
+                warnings.warn(
```
FedericoTartarini: See my previous comment about code repetitions.

FedericoTartarini (changes requested):
Great work, I really like what you have done. Please review all my comments in line and let me know what you think. Please also check that you have modified all the functions/models in pythermalcomfort. Have you included the new modification to the valid range function to all the models in PyDermalComfort?

pythermalcomfort/models/pmv_ppd_iso.py
JitaoFu: Thanks for the push on this — I dug into it carefully and agree the current `_check_standard_compliance_array` design is mixing concerns. Below is the analysis and the refactor I'd like to propose.

---

### Why refactor (not just delete)

`_check_standard_compliance_array` is doing two different jobs:

1. **Single-variable applicability limits** — e.g. `tdb` in `[10, 40]`, `met` in `[1.0, 4.0]`. These are independent per-variable checks, functionally equivalent to `valid_range`, just bundled.
2. **Cross-variable compliance rules** — the ASHRAE 55 `airspeed_control` block: complex multi-variable conditions (`v > 0.8 & clo < 0.7 & met < 1.3`), a call to `operative_tmp` to derive `to`, three branches with standard-specific warning text. This can't be expressed as `valid_range` calls.

Mixing job 1 and job 2 in one function is what creates the inconsistency — single-variable checks duplicated across both `valid_range` and `_check_standard_compliance_array`, with no clear rule for when to use which.

I checked actual usage across the codebase:

| Standard | Models that use it | Cross-variable rules? |
|----------|-------------------|-----------------------|
| ASHRAE 55-2023 | 6 (`pmv_ppd_ashrae`, `adaptive_ashrae`, `set_tmp`, `ankle_draft`, `vertical_tmp_grad_ppd`, `use_fans_heatwaves`) | ✅ (`airspeed_control`) |
| ISO 7730-2005 | 1 (`pmv_ppd_iso`) | ❌ |
| ISO 7933-2004 / 2023 | 1 (`phs`) | ❌ |
| `fan_heatwaves` | 1 (`use_fans_heatwaves`) | ❌ |

ASHRAE 55 is the only standard that is both shared across many models and has irreducible multi-variable logic. Everything else is a single-use bundle of single-variable limits.

---

### Proposed structure

**Keep one focused helper for ASHRAE 55 in `utilities.py`**, renamed to make its scope explicit:

```python
def _check_ashrae55_compliance(tdb, tr, v, met=None, clo=None,
                                airspeed_control=True, v_limited=None):
    """ASHRAE 55-2023 input applicability check.
    Table 7.3.4 single-variable limits + §7.2.1.2 airspeed_control rules."""
    ...
```

Justified because: 6 models share it, and the `airspeed_control` block (~50 lines of cross-variable logic with standard-specific warnings) genuinely needs a single home.

**Inline the other three standards into their respective model files**, following the style already established in `utci.py`:

```python
# pmv_ppd_iso.py
if limit_inputs:
    tdb_valid = valid_range((10.0, 30.0), tdb=tdb)
    tr_valid  = valid_range((10.0, 40.0), tr=tr)
    v_valid   = valid_range((0.0, 1.0),   v=vr)
    met_valid = valid_range((0.8, 4.0),   met=met)
    clo_valid = valid_range((0.0, 2.0),   clo=clo)
    pmv_valid = valid_range((-2, 2),      pmv=pmv_array)
    all_valid = ~(np.isnan(tdb_valid) | np.isnan(tr_valid) | np.isnan(v_valid)
                  | np.isnan(met_valid) | np.isnan(clo_valid) | np.isnan(pmv_valid))
    pmv_array = np.where(all_valid, pmv_array, np.nan)
```

Justified because: each is used by exactly one model, no cross-variable logic, the limit values stay visible at the use site, and it matches `utci.py`'s existing style.

---

### Why I'm not abstracting these into private `_check_xxx_compliance` functions

I considered this and decided against it. For a 6–8 line block with no branching, wrapping it in a private function would:

- Hide the limit values behind a function call (you'd have to jump to the top of the module to see what range ISO 7730 actually uses for `tdb`)
- Add naming and signature overhead for code that's used exactly once

The rule I'd commit to: **abstract into a separate function only when (a) shared across multiple models, (b) involves cross-variable logic, or (c) exceeds ~15–20 lines.** ASHRAE 55 hits (a) , (b) and (c); the others hit none.

---

### Summary

After this refactor, the codebase follows a single principle:

- **Shared + complex** → centralised helper (`_check_ashrae55_compliance`)
- **Single-use + single-variable** → inlined in the model file with `valid_range`

This gives every model exactly one style of compliance block, and the choice of style is determined by an objective criterion rather than historical accident. 

Let me know if you're happy with this direction.

Click link to go back the summary comment below https://github.com/pythermalcomfort/pythermalcomfort/pull/315#issuecomment-4466548922

pythermalcomfort/models/pmv_ppd_iso.py
JitaoFu: Thank you for pointing it out. I initially treated pmv_array as an output rather than an input, so I left it without a name. But the code is already setting results to NaN when PMV falls outside [-2, +2] — which means it's effectively an applicability boundary in the ISO 7730 sense. Without a warning, users would see NaN with no explanation. 

I'll keep the behaviour consistent with the other checks.

pythermalcomfort/models/adaptive_ashrae.py
JitaoFu: Got it, will look into it

pythermalcomfort/models/ankle_draft.py
JitaoFu: I'm not sure about the original design intent, but from what I've observed, functions that have applicability limit checks generally include a limit_inputs parameter. Happy to add it here if that's the expected behaviour.

pythermalcomfort/models/vertical_tmp_grad_ppd.py
JitaoFu: Got it, will add it

pythermalcomfort/shared_functions.py
JitaoFu: Understood, pre-commits has been installed.

pythermalcomfort/shared_functions.py
JitaoFu: Yes, I agree, it's worth to do and these changes are straightforward to implement.

pythermalcomfort/shared_functions.py
JitaoFu: Thanks for pointing this out, I wasn’t aware of this convention.

pythermalcomfort/utilities.py
JitaoFu: Got it.

JitaoFu:
Hi Federico,

Thanks for the detailed review. Before pushing any code changes, I'd like your decision on the two architectural questions below (points 1 and 2). Points 3 and 4 are procedural follow-ups that will be handled mechanically once 1 and 2 are settled.

---

### 1. `valid_range` signature (re: `adaptive_ashrae.py:134`, `shared_functions.py:8`)

You asked whether we can drop the manual `param_name="tdb"` string and have `valid_range` extract the variable name automatically. Two approaches:

**Option A — frame inspection**

Use `inspect`/`ast` to read the caller's source line and parse the variable name from the call AST at runtime.

```python
# Definition (conceptual)
def valid_range(x, valid, param_name=None) -> np.ndarray:
    if param_name is None:
        param_name = _extract_caller_argname()  # inspect frame
    ...

# Call site — unchanged from current style
valid_range(tdb, (10, 40))
```

Pros: call sites look identical to the current code, no migration needed.

Cons:
- Still needs a `param_name` fallback for expressions (`tr - tdb`), subscripts (`params["tdb"] in check_compliance_array`), and literals in tests, even if we don't count the calls in `check_compliance_array` ~5% of call sites will have to pass the `param_name` like `valid_range(tdb, (10, 40), 'tdb' )`
- Fragile: Frame inspection is inherently fragile and especially unstable in Jupyter environments.

**Option B — kwargs**

Change the signature so the variable name comes in as a keyword-argument key.

```python
# Definition
def valid_range(valid, **kwargs) -> np.ndarray:
    if len(kwargs) != 1:
        raise TypeError("valid_range expects exactly one keyword argument")
    ((param_name, x),) = kwargs.items()
    ...

# Call site
valid_range((10, 40), tdb=tdb)
```

Pros:
- 100% coverage — works identically for expressions, subscripts, and literals
- Pure Python kwargs mechanism — statically analysable, no runtime introspection or fragility
- Eliminates `param_name` entirely; no fallback needed

Cons:
- `**kwargs` enforcing exactly one key is an unusual API shape — mitigated by a clear docstring and a runtime check
- All ~20 call sites in `utilities.py` (and `pmv_ppd_iso.py:187`) need to be updated

**Recommendation: Option B.** Frame inspection looks cleaner at the call site, but it trades that visual nicety for runtime fragility and an unavoidable fallback parameter. The kwargs form is unusual but clean and stable.

---

### 2. `_check_standard_compliance_array` (re: `pmv_ppd_iso.py:179`, `utilities.py:453`)
Please see the details via https://github.com/pythermalcomfort/pythermalcomfort/pull/315#discussion_r3252551957


---

### 3. `valid_range` internal refactor (re: `shared_functions.py:16`, `:25`, `:26`)

Procedural follow-up to point 1. I've understood your suggestions — define `msg` as a variable first, use a single `warnings.warn` call covering both scalar and array branches, and replace `"some values"` with the actual count. I'll apply these together with whichever signature option you pick in point 1, so the diff stays coherent for re-review.

---

### 4. `limit_inputs` follow-up

Procedural follow-up. The missing `limit_inputs` parameters on `ankle_draft` and `vertical_tmp_grad_ppd` (and the related moves into the `if limit_inputs:` block) will be added as a final commit after the structural changes above are in place.

---

Looking forward to your decision on points 1 and 2.

Thanks,
Jitao

FedericoTartarini:
1. I don't like the option B with the key value arguments. I like option A, As per your suggestion we can have that the parameter name is optional and if it is optional we are going to get the caller argument name. However we need to check that this doesn't break the code on notebooks.
2. I like your suggestion, you can create _check_ashrae55_compliance and use Inline the other three standards into their respective model files

JitaoFu:
Hi @FedericoTartarini, I've pushed commit `1dc4ce3` with the structural changes following your decisions on points 1 and 2. This is the core review-response commit; `limit_inputs` (point 4) will follow as a separate commit so each is easier to review.

---

## What's in this commit

### 1. `valid_range` — Option A (frame inspection)

- `param_name` is now optional; when omitted, the caller's argument name is auto-extracted via `_extract_caller_argname()` (frame + AST inspection)
- 29 of 36 call sites pass a bare variable (`valid_range(tdb, (10, 40))`) and work unchanged with no migration. The remaining 7 explicit calls are: 6 subscript calls inside `_check_ashrae55_compliance` (frame inspection can't recover names from `params["v"]`-style accesses), and one expression call in `utci.py:118` (`param_name="tr - tdb"`)
- Internally also addresses points 8–10: `msg` is now built first then passed to a single `warnings.warn`, `"some values"` is replaced by the actual count + values + indices, and scalar/array branches share one warning template

### 2. `_check_standard_compliance_array` split (point 2)

- Renamed the kept helper to `_check_ashrae55_compliance` — only the ASHRAE 55 branch + `airspeed_control` cross-variable rules remain in `utilities.py`
- ISO 7730-2005, ISO 7933-2004/2023, and `fan_heatwaves` checks are inlined into `pmv_ppd_iso.py`, `phs.py`, and `use_fans_heatwaves.py` respectively, each using `valid_range` directly
- The three `airspeed_control` warnings now share a `_format_violation_detail` helper so their text matches the unified `valid_range` format

### 3. Airspeed variable naming consistency

- `_check_ashrae55_compliance` now accepts a `v_param_name` kwarg (default `"v"`). Three callers whose signature names the airspeed argument `vr` (`pmv_ppd_ashrae`, `ankle_draft`, `vertical_tmp_grad_ppd`) pass `v_param_name="vr"`, so warnings emitted on their behalf use the variable name the user actually typed

### 4. Warning text changes

| | Format |
|---|---|
| Old | `"Some values of 'v' are outside the applicability limits..."` |
| New (scalar) | `"'v' has value X outside the applicability limits [lo, hi] and will be set to NaN."` |
| New (array) | `"'v' has N values [...] at indices [...] outside the applicability limits [lo, hi] and will be set to NaN."` |

All `pytest.warns(match=...)` assertions have been updated.

---

## Verification

- Full test suite: **323 passed, 1 xfailed** (matches `development` baseline)
- `ruff check`, `ruff format`, `docformatter` clean on touched files
- pre-commit checked
- Jupyter / IPython smoke check (your prerequisite for Option A): variable-name extraction works for top-level cell calls, expressions, subscripts, and explicit `param_name=` — 7/7 PASS

---

## Next

`limit_inputs` for `ankle_draft` / `vertical_tmp_grad_ppd` (points 2, 5, 6) and the associated `classes_input.py` + docstring updates will come as the next commit. Happy to wait for your feedback on this one first.

JitaoFu:
Hi Federico, I've pushed `22ce87e` — the `limit_inputs` follow-up (point 4 from the review thread).

---

## Changes

- **`ankle_draft` / `vertical_tmp_grad_ppd`**: added `limit_inputs: bool = True`. The ASHRAE 55 compliance check and NaN-mask are now inside `if limit_inputs:`. For `ankle_draft`, the `vr < 0.2` equation-applicability `ValueError` is also gated by the flag, matching how `limit_inputs=False` works in `pmv_ppd_ashrae` / `adaptive_ashrae`.

- **`vertical_tmp_grad_ppd`**: renamed the 5-tuple unpacking from `(tdb_valid, tr_valid, v_valid, met_valid, clo_valid)` to `(tdb_valid, tr_valid, met_valid, clo_valid, v_limited_valid)` to match the helper's actual return order (no `v=` is passed here). **Naming-only fix — the mask combines the same five arrays as before.**

- **`classes_input.py`**: threaded `limit_inputs` through `AnkleDraftInputs` / `VerticalTGradPPDInputs` `__init__`.

- **Docstrings**: added a `limit_inputs` parameter description to both models.

---

## Verification

- 328 passed, 1 xfailed (5 new tests covering `limit_inputs=False`); ruff / docformatter clean.

This should close out the review thread. Let me know if anything needs adjusting.

FedericoTartarini (approved):

PR #316, opened 2026-05-10
fix: restrict compliance attribute to ASHRAE model only

### LINKED ISSUE #296: pmv_ppd_iso compliance
The PMVPPD return class now includes an attribute called compliance which should only be returned for the PMVPPD model for the ASHRAE However, this is also returned for all the other PMVPPD models and it is very confusing because if the results for the ISO model are within the neutral range still the compliance says "none" which is the default value so I need to fix that.

JitaoFu:
## What this PR does

- Introduce `PMVPPDAshrae` as a subclass of `PMVPPD` that carries the `compliance` field
- `pmv_ppd_ashrae` now returns `PMVPPDAshrae` (with `compliance`); `pmv_ppd_iso` returns the base `PMVPPD` (no `compliance` attribute)
- Remove the confusing `compliance: None` default from the base `PMVPPD` dataclass
- Update docs to reference `PMVPPDAshrae` instead of `PMVPPD` for the ASHRAE model

## Why this change is needed

Previously `PMVPPD` carried a `compliance` field defaulting to `None`, which was returned by `pmv_ppd_iso` even though `compliance` is an ASHRAE-only concept. This was misleading — users accessing `.compliance` on an ISO result would silently get `None` with no indication that the field does not apply. The new subclass design makes the type system reflect the actual semantics.

## How I tested it

- Added `test_no_compliance_attribute` to `test_pmv_ppd_iso.py` to assert that ISO results do not expose a `compliance` attribute, preventing regression
- Ran the full test suite

## Notes for reviewers

- This is a **breaking change** for any code that type-checks the return value of `pmv_ppd_ashrae` as `PMVPPD` — it now returns the more specific `PMVPPDAshrae`. Code that only accesses `.pmv`, `.ppd`, `.tsv` is unaffected since `PMVPPDAshrae` inherits from `PMVPPD`. 
- `compliance` is no longer optional (`None`) — it is always present and always a `bool | list[bool]` on ASHRAE results unless the compliance check failed, all fields in return will be `nan`.

Closes #296


TIAN-TOM:
@JitaoFu same base-branch issue here as #315. Could you retarget this PR from `master` to `development` as well, so CI and CodeRabbit can run?

Also, the PR description says `compliance` is always `bool` or `list[bool]`. I think that needs a small tweak: when `limit_inputs=True` and an input is out of range, scalar calls can return `nan`, and vector calls can include `nan` in an object-dtype array. Could you update the description, and check whether the dataclass annotation should mention or document that case?

pythermalcomfort/models/pmv_ppd_ashrae.py
```diff
@@ -28,7 +28,7 @@ def pmv_ppd_ashrae(
     limit_inputs: bool = True,
     airspeed_control: bool = True,
     round_output: bool = True,
-) -> PMVPPD:
+) -> PMVPPDAshrae:
```
TIAN-TOM, line 31: This return annotation was updated, but the `Returns` section lower in the same docstring still names `PMVPPD` and links to `PMVPPD`. Could you update that docstring to `PMVPPDAshrae` as well?

JitaoFu:
Thanks for the feedback, @TIAN-TOM
- The base-branch has been retargeted to `development` 
- PMVPPDAshrae.compliance docstring has been updated to document that when `limit_inputs=True` and an input is outside the standard applicability limits, compliance returns `nan`. As for the type annotation, I've kept it as `bool | list[bool]` since the `nan` case is an edge condition already covered by the doc note. 
- The docstring issue has been resolved.
- The PR description has been changed as well.

FedericoTartarini (approved):

PR #331, opened 2026-05-18
feat: add set_adaptive_regions to SummaryPlot for adaptive method classification

qdou0670:
## Add adaptive comfort classification to SummaryPlot

This PR adds `set_adaptive_regions()` to `SummaryPlot`, enabling it to classify DataFrame rows using the adaptive model's `acceptability_*` outputs instead of fixed thresholds.

### Why

`SummaryPlot` previously only supported fixed-threshold classification via `set_regions()` (e.g., PMV thresholds), while the adaptive models already return per-row `acceptability_80/90` (ASHRAE) and `acceptability_cat_i/ii/iii` (EN) boolean outputs, which are the correct way to classify data.

#### `_shared.py`
- Added `outside_label` and `outside_color` to `_PlotDefaults.Summary`

#### `adaptive.py`
- Added `_ACCEPTABILITY_FIELDS` dict mapping standard names to model output field names

#### `summary.py`
- Added `_compute_adaptive_percentages()` — classifies rows using model `acceptability_*` outputs
- Added `SummaryPlot.set_adaptive_regions()` — configures adaptive classification mode
- Modified `SummaryPlot.__init__()` — added adaptive mode state fields
- Modified `SummaryPlot.set_regions()` — clears adaptive state when called (mutual exclusion)
- Modified `SummaryPlot.plot()` — supports both threshold and adaptive modes

### Design decisions

- **`set_adaptive_regions()` instead of overloading `set_regions()`** — The two modes have completely different parameters (`output`/`thresholds` vs `model_func`/column mappings). Merging them into one method would create 10+ parameters, most mutually exclusive. A separate method makes the interface clear and self-documenting.

- **Mutual exclusion** — Calling `set_regions()` clears adaptive state; calling `set_adaptive_regions()` clears threshold state. Only one mode is active at a time.

- **Reuses existing rendering** — `plot()` delegates to the same `_prepare_axis()` and `_plot_summary()` functions regardless of mode. The only difference is how `percentages`, `labels`, and `colors` are computed.

- **Shares `RegionsConfig` with `AdaptivePlot`** — The same config object controls both the comfort band chart and the summary bar, guaranteeing consistent labels and colors.

### Potential issue

The code uses four parameters—`tdb`, `tr`, `t_running_mean`, and `v`—to calculate the point's location. However, in CBE, the point's location remains unchanged regardless of the chosen `v` value. Yet, in the `operative_tmp` of `utilities.py`, `v` is indeed involved in the calculation, meaning `v` affects the point's location. Therefore, I assume our `utilities.py` are correct and the CBE display is incorrect.

This leads to different AdaptivePlots for points with different `v` values. Therefore, while overlaying all points onto a single AdaptivePlot results in a correct statistical panel, the visual display of the overlay is incorrect. This is unless the `v` values ​​of all points and the `v` value of the AdaptivePlot fall within the same interval in both `operative_tmp` and `adaptive_cooling_effect`. The simplest case is when the `v` values ​​of all points are the same as those in the AdaptivePlot.

### Usage

```python
import matplotlib.pyplot as plt
import numpy as np
import pandas as pd

from pythermalcomfort.models import adaptive_ashrae, adaptive_en
from pythermalcomfort.plots.matplotlib import AdaptivePlot, SummaryPlot
from pythermalcomfort.utilities import operative_tmp

# 1. Points have different v, and therefore cannot be plotted on one single AdaptivePlot.

np.random.seed(42)
n = 10
df = pd.DataFrame({
    "tdb": np.random.uniform(18, 32, n),
    "tr": np.random.uniform(18, 32, n),
    "t_running_mean": np.random.uniform(10, 33, n),
    "v": np.random.uniform(0.05, 0.5, n),
})

fig, (ax1, ax2) = plt.subplots(2, 1, figsize=(8, 7))

AdaptivePlot(adaptive_ashrae).plot(ax=ax1, title="Adaptive Comfort (ASHRAE 55)")

SummaryPlot(df).set_adaptive_regions(
    model_func=adaptive_ashrae,
    tdb="tdb", tr="tr", t_running_mean="t_running_mean", v="v",
).plot(ax=ax2, title="Comfort Distribution")

fig.tight_layout()
plt.show()

# 2. Points have the same v, and therefore can be plotted on the same AdaptivePlot.

np.random.seed(42)
n = 10
df = pd.DataFrame({
    "tdb": np.random.uniform(18, 32, n),
    "tr": np.random.uniform(18, 32, n),
    "t_running_mean": np.random.uniform(10, 33, n),
    "v": 1.2,
})

fig, (ax1, ax2) = plt.subplots(2, 1, figsize=(8, 7))

AdaptivePlot(adaptive_ashrae).set_params(v=1.2).plot(ax=ax1,title="Adaptive Comfort (ASHRAE 55)")

to = operative_tmp(df["tdb"].values, df["tr"].values, df["v"].values, standard="ashrae")
ax1.scatter(df["t_running_mean"], to, c="black", s=5)

SummaryPlot(df).set_adaptive_regions(
    model_func=adaptive_ashrae,
    tdb="tdb",
    tr="tr",
    t_running_mean="t_running_mean",
    v="v",
).plot(ax=ax2, title="Comfort Distribution")

fig.tight_layout()
plt.show()
```
adaptive_summary
adaptive_summary_vfixed

### DIFF
--- a/pythermalcomfort/plots/matplotlib/_shared.py
+++ b/pythermalcomfort/plots/matplotlib/_shared.py
@@ -142,6 +142,8 @@ class _PlotDefaults:
         v_bar_x: float = 0.0
         v_bar_width: float = 0.80  # bar fills most of the horizontal span
         v_label_x_offset: float = 0.42  # just past bar right edge (0.40)
+        outside_label: str = "Outside"
+        outside_color: str = "#D9D9D9"
 
 
 # ── package-wide Matplotlib style ─────────────────────────────────────────
--- a/pythermalcomfort/plots/matplotlib/adaptive.py
+++ b/pythermalcomfort/plots/matplotlib/adaptive.py
@@ -79,6 +79,11 @@ _BAND_KEYS: dict[str, list[str]] = {
     std: [b.key for b in cfg["bands"]] for std, cfg in _STANDARD_CONFIGS.items()
 }
 
+_ACCEPTABILITY_FIELDS: dict[str, list[str]] = {
+    "ashrae": ["acceptability_80", "acceptability_90"],
+    "en": ["acceptability_cat_iii", "acceptability_cat_ii", "acceptability_cat_i"],
+}
+
 
 # ── public config ──────────────────────────────────────────────────────────
 
--- a/pythermalcomfort/plots/matplotlib/summary.py
+++ b/pythermalcomfort/plots/matplotlib/summary.py
@@ -21,6 +21,13 @@ from pythermalcomfort.plots.matplotlib._shared import (
     _is_light_color,
     _PlotDefaults,
 )
+from pythermalcomfort.plots.matplotlib.adaptive import (
+    _ACCEPTABILITY_FIELDS,
+    _MODEL_TO_STANDARD,
+    _STANDARD_CONFIGS,
+    RegionsConfig,
+    _BandSpec,
+)
 
 # ── result container ───────────────────────────────────────────────────────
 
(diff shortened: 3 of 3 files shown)


stefanoschiavon:
@FedericoTartarini @t-kramer @qdou0670 
Here is a polished GitHub comment:

I have a few suggestions to improve the clarity and visual consistency of the adaptive comfort figures. 
I would remove all grid lines, both horizontal and vertical, since they add visual noise without much benefit. 
I would also consider removing the black box around the figure, or making it a very light gray, because the current black outline is too prominent. 
The number of y-axis labels could be reduced to a maximum of five to make the figure cleaner.

There also seems to be an issue with visual weight. The bar chart is too large (y-axis) for the amount of information it conveys. I would make the bars thinner.
I would create a larger gap between the first and second graphs.

I would also consider removing the titles, such as “Adaptive Comfort R55” and “Comfort Distribution”. 
The gray area for the "outside the comfort zone" could be made white (light gray outline around it), to improve consistency with the other comfort regions in the figure above.

FedericoTartarini:
Closing this as superseded by SummaryPlot.set_categories(...). Adaptive models can calculate their acceptability categories first, then SummaryPlot can render those categories directly. This keeps the calculation explicit and avoids adding a second, model-specific configuration mode to SummaryPlot.

Thank you for the implementation and the useful exploration of adaptive summary plots.

PR #337, opened 2026-05-28
Add `HeatmapPlot` for temporal comfort visualisation

t-kramer:
Adds a new `HeatmapPlot` class to `pythermalcomfort/plots/matplotlib/` that displays hourly thermal comfort model output on a 2-D temporal grid: hours of the day (y-axis, 0 at bottom) against days of the year (x-axis). Each cell is coloured by the threshold region it falls in; NaN cells (e.g. out-of-model-range hours) appear blank.

## Key design points

- Follows the same fluent API and `set_regions()` interface as `SummaryPlot` and `ThresholdPlot`, so region thresholds, labels, and colours are shared across plot types
- Accepts a DataFrame with a `DatetimeIndex`; NaN values in the output column are permitted and shown as blank cells
- `x_label_format` controls whether the x-axis shows abbreviated month names (default) or numeric day-of-year ticks
- Returns a `HeatmapPlotResult` with direct access to the `pcolormesh` artist for further customisation

## Files changed

- `pythermalcomfort/plots/matplotlib/heatmap.py` — new module
- `pythermalcomfort/plots/matplotlib/_shared.py` — added `_PlotDefaults.Heatmap`
- `pythermalcomfort/plots/matplotlib/__init__.py` — exported `HeatmapPlot`, `HeatmapPlotResult`
- `docs/documentation/plots/matplotlib/heatmap_plot.ipynb` — example notebook covering PMV (office building), Adaptive ASHRAE (acceptability bands), UTCI (outdoor stress categories), and Heat Index (subtropical climate), each paired with a `SummaryPlot`

--- 

**Examples**

Screenshot 2026-05-28 at 4 02 13 PM

Screenshot 2026-05-28 at 4 02 27 PM

Screenshot 2026-05-28 at 4 02 36 PM

Screenshot 2026-05-28 at 4 02 45 PM


stefanoschiavon:
@t-kramer, rhe heatmaps look very nice. Overall, the quality of the figures is very good. I have a few suggestions that may further improve readability and consistency.

- [ ] the month labels are somewhat unclear because they appear to indicate the center of each month rather than the full range. I am not sure this is critical, but you may want to test how the figure looks if the beginning and end of each month are marked, with the month name centered between them. This would increase the number of ticks, but it would make it clearer when each month starts and ends.
 
- [ ] please check capitalization. In most labels, the second word is not capitalized, but “Hour of Day” is. I would consider changing it to “Hour of day” for consistency.

- [ ] I suggest adding secondary marks on the y-axis at 3, 9, 15, and 21 hours.

- [ ] I would remove the black border around the heatmaps. One option is to remove the border entirely. Another option is to use a light gray border instead.

- [ ] in some cases the bar chart on the side seems too far from the main heatmap, for example in the UTCI figure. Please check why this is happening. Is there a way to not plot the bar line when it is very low, sometimes is so small that seems a printing error.

FedericoTartarini:
Thank @t-kramer for working on this. The temporal comfort heatmap is a useful visualisation, and the examples clearly show the value of pairing it with a summary of time spent in each category.

However, I do not think we should be adding a HeatmapPlot as a public plotting class at this time and the pull request needed to be updated since it now conflicts with the latest version of the code. The proposed code in HeatmapPlot is short and works well as ordinary Matplotlib code. A public class would also require us to maintain choices around aggregation, missing timestamps, leap years, time zones, daylight-saving transitions, partial years, and multi-year data. 

I have, therefore, added an example in the documentation called “Annual Comfort Heatmap + Summary”. It shows how to create the heatmap with standard Matplotlib and pair it with SummaryPlot, using the same category definitions and colours for both charts. What do you think? Happy to hear your thoughts and happy for you to work on the example and improve it and implement the suggestions from Stefano. If on the other hand you think we really need this class we can consider adding it but starting from the latest code in`development`.

t-kramer:
@FedericoTartarini To be honest, I think this would be one of the most useful ones for many users... I'm not saying this because I wrote the code for this example, but because of my own experience building this from scratch a few times. I would have loved to just use a function within `pythermalcomfort` back then to streamline the process of dealing with the exact same points you list above. But I'll leave the decision to you, documentation is fine for me. I can also add the examples to our CBE plot library.

FedericoTartarini:
Could you please have a look at the example that I have uploaded in the development branch as a Jupyter Notebook, it is only a few lines of code. However if you think that this chart is needed, we can add it, but we need to take into account all the issues that I have mentioned before:
- aggregation, 
- missing timestamps, 
- leap years, 
- time zones, 
- daylight-saving transitions, 
- partial years 
- multi-year data.
- timestamp formatting of the index
- handling of different timestamps

Moreover, I will create a class only for the heatmap and plot the summary as a separate plot. Let me know your thoughts

t-kramer:
@FedericoTartarini I just checked it, looks good to me!

I think the issues you point out are definitely justified. But at the same time, handling those would help the users and streamlines the process of creating a graphic like this. Anyway, I think you should decide and having it as an example is great 👍🏼

FedericoTartarini:
Let's design a simple, easy-to-maintain wrapper for the HeatMap component if it makes sense. To make sure it stays maintainable, we should first look at how the existing data is structured and reuse patterns we already have in our plotting tools. By keeping things clear and focused, we can build something flexible and reusable. This will reduce confusion, make the code easier to read, and simplify future updates. Let's start by opening a dedicated issue to outline the basic design, how data flows through it, and how it connects with the rest of the visualization system. Once we have agreed on the structure of the dataclass we can then write the code. I really like the grilling skill from mattpokok which really question me when i want to implement somethign.

PR #345, opened 2026-06-22
feat: extend sports heat stress extreme risk scale to 4.9

yehui-h:
## Summary
- Extend `sports_heat_stress_risk` extreme risk interpolation from a fixed `4.0` to `4.0-4.9`.
- Keep `43.5°C` as the maximum high-risk threshold, while allowing lower model-derived `t_extreme` values to start the extreme range earlier.
- Update return docs and tests to reflect the new `1.0-4.9` risk scale.

## Versioning
- This PR only deploys the feature change.
- The `development` version appears to be out of sync with `master`, so no version bump is included for now.
- There is currently no related code change in this PR to bump the package version.

## Testing
- `ruff check pythermalcomfort/models/sports_heat_stress_risk.py pythermalcomfort/classes_return.py tests/test_sports_heat_stress_risk.py`
- `pytest -q tests/test_sports_heat_stress_risk.py`

### DIFF
--- a/pythermalcomfort/classes_return.py
+++ b/pythermalcomfort/classes_return.py
@@ -1077,8 +1077,9 @@ class SportsHeatStressRisk(AutoStrMixin):
     Attributes
     ----------
     risk_level_interpolated : float or list of floats
-        Interpolated risk level (1.0-4.0), [dimensionless].
-        Risk levels: 1-<2 = low, 2-<3 = moderate, 3-<4 = high, 4 = extreme.
+        Interpolated risk level (1.0-4.9), [dimensionless].
+        Risk levels: 1-<2 = low, 2-<3 = moderate, 3-<4 = high,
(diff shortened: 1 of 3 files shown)


FedericoTartarini:
Previously, the extreme risk interpolation (4.0 → 4.9) was anchored to a fixed 50°C ceiling regardless of humidity. This meant the width of the extreme range varied depending on where t_extreme fell — narrower at high humidity, wider at low humidity.                                             
                                                                                                       
  The change replaces that fixed anchor with a dynamic one: t_extreme + 5°C. The extreme risk range now
  always spans exactly 5°C above the humidity-dependent extreme threshold, reaching 4.9 at that point  
  and capping there. This makes the scale consistent across conditions — 5°C into the extreme zone     
  always means maximum risk, regardless of how hot or humid it is.

FedericoTartarini (approved):

PR #347, opened 2026-06-23
Implement shoen model and update related documentation

stefanoschiavon:


pythermalcomfort/models/heat_index_schoen.py
```diff
@@ -0,0 +1,71 @@
...
+        To access the `hi` value, use the `hi` attribute of the returned `HI` instance, e.g., `result.hi`.
+
+    Examples
+    --------
+    .. code-block:: python
+
+        from pythermalcomfort.models import heat_index_schoen
+
+        result = heat_index_schoen(tdb=29, rh=50)
+        print(result.hi)  # todo: add expected output
```
FedericoTartarini: missing value

FedericoTartarini (dismissed):
perfect, but add missing value

FedericoTartarini:
Hi @stefanoschiavon — great work on the Schoen model! One issue I'd like your input on before we merge.

## Stress category lower bound

The current implementation uses `-1000` as the lower bound for the `"no risk"` category:

```python
heat_index_categories = {
    -1000: "no risk",
    32.0: "caution",
    ...
}
```

Because of how `mapping()` works with `np.digitize`, this means **any heat index between -1000°C and 32°C gets labelled `"caution"`** — including values like 5°C or 20°C where there is clearly no heat stress. A quick example:

```python
result = heat_index_schoen(tdb=10, rh=50)
result.stress_category  # returns "caution" — incorrect
```

I can fix this, but I want your input on the right approach. Two options:

### Option A — Set a lower limit (like Rothfusz does)
Use `27.0` as the first key (matching `heat_index_rothfusz`) so values ≤ 27°C return `"no risk"`, and apply `valid_range(tdb, (27.0, np.inf))` to return `NaN` for inputs below the model's applicability range. Users would need to opt out via `limit_inputs=False` if they want values outside the range.

### Option B — Raise a warning for cold conditions
Keep computing a result for any temperature but emit a `UserWarning` when the heat index is clearly outside the heat-stress regime (e.g. below 27°C), suggesting the user consider a cold-stress model instead. The `stress_category` would still return `"no risk"` (with a sensible lower bound like `27.0`), but users would be informed they may be in a cold-stress situation.

My preference would be **Option A** for consistency with the rest of the library, but happy to go with Option B if you think the warning message adds useful context.

Let me know how you'd like to proceed and I'll implement the fix. — @FedericoTartarini

stefanoschiavon:
I agree with Option A. Thank you for finalizing it.

FedericoTartarini (approved):
Approving after the review-feedback fixes (stray file removed, redundant .gitignore line fixed, docstring corrected, shared HI stress-category thresholds, and the -1000 vs 27 no-risk floor bug fixed).

PR #349, opened 2026-07-17
feat: add Croquet sports heat stress preset

FedericoTartarini:
## Summary
- Add `Sports.CROQUET` preset (`clo=0.7, met=4.5, vr=0.5, duration=90`) to `sports_heat_stress_risk.py`
- Add corresponding unit test and registration in the all-sports test

Based on yehui-h's `feature/add_croquet_to_sports` branch, with the comparison plotting script and generated PNGs intentionally left out to keep the PR scoped to the model addition.

## Test plan
- [x] `pytest tests/test_sports_heat_stress_risk.py` — 42 passed, 1 pre-existing unrelated failure (`test_sports_heat_stress_risk_extreme_interpolation`, reproduces identically on `development` without this change; solver-tolerance sensitivity, not caused by this PR)

### DIFF
--- a/pythermalcomfort/models/sports_heat_stress_risk.py
+++ b/pythermalcomfort/models/sports_heat_stress_risk.py
@@ -58,6 +58,7 @@ class Sports:
     BOWLS = _SportsValues(clo=0.5, met=5.0, vr=0.5, duration=180)
     CANOEING = _SportsValues(clo=0.6, met=7.5, vr=2.0, duration=60)
     CRICKET = _SportsValues(clo=0.7, met=6.0, vr=0.75, duration=120)
+    CROQUET = _SportsValues(clo=0.7, met=4.5, vr=0.5, duration=90)
     CYCLING = _SportsValues(clo=0.4, met=7.0, vr=3.0, duration=60)
     EQUESTRIAN = _SportsValues(clo=0.9, met=7.4, vr=3.0, duration=60)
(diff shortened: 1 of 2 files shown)


FedericoTartarini:
Investigated the CI failure — it's **pre-existing on `development` and unrelated to this PR's diff**.

The only failing test is `test_sports_heat_stress_risk_extreme_interpolation`, which exercises `Sports.FISHING` (not Croquet) and asserts `risk_level_interpolated == 4.3` at `tdb=45`; CI gets `4.2`.

I reproduced the identical failure on `origin/development` HEAD (`fc5f30e`) with none of this PR's changes applied, and confirmed `fc5f30e`'s own CI run already shows this same test failing on `test (3.10, ubuntu-latest)`. So the regression was introduced by `fc5f30e` ("fix: scale extreme risk segment so 4.9 is reached at exactly +5°C"), two commits before this PR, not by the Croquet preset addition.

Root cause: in `sports_heat_stress_risk.py`, `extreme_entry_t` is derived from the raw (unrounded) `brentq`-solved threshold, but the test's expected value assumes the rounded `43.5°C` threshold that's actually returned to callers. That small delta, combined with the result being **floored** (not rounded) to 1 decimal, pushes borderline values like this one down a bucket (4.26x → floors to 4.2 instead of rounding to 4.3).

Merging with this pre-existing, unrelated failure noted. Will fix the underlying rounding/truncation bug on `development` in a follow-up.

PR #399, opened 2026-09-13
docs: clarify air speed assumptions for AdaptivePlot scatter overlays

### LINKED ISSUE #366: Adaptive comfort: air speed (v) affects operative_tmp point placement but AdaptivePlot renders a single fixed-v comfort zone
## Summary

Flagged by a contributor (qdou0670) in the description of PR #331 (`set_adaptive_regions` for `SummaryPlot`), and worth tracking independently since it affects correctness of overlaying scatter data onto `AdaptivePlot`, not just the new SummaryPlot feature.

## Issue

`utilities.operative_tmp()` computes operative temperature using `tdb`, `tr`, and `v` — `v` genuinely shifts where a given data point lands on the psychrometric/operative-temperature axis (via the SET-based adjustment for air speed).

However, `AdaptivePlot` renders the ASHRAE-55 / EN 16798 comfort bands for a **single fixed `v`** (default or via `set_params(v=...)`), via `adaptive_cooling_effect`. If a user overlays a scatter of points whose `v` values vary row-to-row onto a single `AdaptivePlot`, each point's true operative temperature was computed with its own `v`, but the comfort band boundaries drawn on the chart assume one fixed `v` for every point. So:

- Aggregate statistics (percent-comfortable style summaries, like the new `SummaryPlot.set_adaptive_regions()`) are computed correctly per-row, using each row's own `v` — this part is fine.
- But the *visual* overlay of raw scatter points on a single `AdaptivePlot` is only strictly accurate when either (a) all points share the same `v` as the plot's configured `v`, or (b) all points and the plot's `v` happen to fall within the same "cooling effect" bracket in both `operative_tmp` and `adaptive_cooling_effect`.

See PR #331's own repro (two examples: heterogeneous `v` vs. fixed `v=1.2` for both the scatter data and the `AdaptivePlot`).

## Suggested follow-up

- Decide whether `AdaptivePlot` should warn (or document clearly) when used to overlay scatter data with non-uniform `v`.
- Consider whether a documentation note / runtime check in `AdaptivePlot.plot()` is warranted when a caller overlays external scatter data with heterogeneous `v`.
- Not a blocker for merging adaptive-mode support in `SummaryPlot` (which classifies per-row using each row's real `v` and doesn't have this issue), but should be documented so users don't misread overlaid scatter plots.

harry11096:
## Summary

AdaptivePlot draws comfort bands using one fixed air speed, so external scatter points measured at different air speeds can appear outside a band even when their per-row model results indicate comfort.

- Document how to choose the chart's `v`, including using a representative mode or median when measurements vary.
- Explain that mixed-air-speed overlays are approximate and identify the per-row acceptability outputs for ASHRAE and EN.
- Add a worked example at `tdb=tr=29`, `t_running_mean=25`, comparing `v=0.1` and `v=1.2`: the 90% upper limit changes from 28.1 to 30.3 °C, although both scatter points occupy the same location.

Documentation only; no calculation, plotting behavior, or runtime warnings changed.

Closes #366.

## Validation

- `pytest -q --tb=short --disable-warnings`: 509 passed, 1 expected failure (Python 3.13).
- `ruff check ./pythermalcomfort ./tests`: passed.
- `ruff format --check ./pythermalcomfort ./tests`: passed.
- Ruff formatting check for the updated notebook: passed.
- Tox was not available locally; the CI test matrix remains to be run.

Student peer review is pending before final maintainer review.


FedericoTartarini (approved):

FedericoTartarini (approved):
Docs-only clarification of AdaptivePlot air-speed assumptions. Worked example numbers verified locally against adaptive_ashrae; CI green.

PR #414, opened 2026-09-18
feat(models): add Liljegren WBGT via optional lwbgt backend

zyf0717:
# Description

Add `wbgt_liljegren()` to estimate outdoor WBGT and component temperatures from weather data. Existing `wbgt()` combines supplied component temperatures; this API performs the underlying calculations through optional [`lwbgt>=0.3.0,<0.4`](https://pypi.org/project/lwbgt/), installed with `pythermalcomfort[lwbgt]`.

Includes array broadcasting, input validation, missing-data handling, solver diagnostics, tests and documentation. Assumptions and active wind adjustment generate one combined INFO message per call; unused wind settings are unlogged at 2 m. Calculation failures return nonzero status and NaN for failed outputs, preserving valid partial results without Python warnings or failure logs. Status `-2` identifies invalid backend output despite native success (or invalid native status). The normal tox environment installs the extra.

Also fixes broken contribution-guide links in the README and installation docs that returned **404 Documentation page not found**.

Related issue: none. Required code-owner reviewer: @FedericoTartarini.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Changes to Core Features:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your core changes, as applicable?
* [x] Have you successfully run tests with your changes locally?

## New Feature Submissions:

1. [x] Does your submission pass tests?
2. [x] Have you lint your code locally before submission?

# How Has This Been Tested?

- [x] Python 3.10, NumPy 2.2.6, lwbgt 0.3.0: **799 passed, 1 existing PHS expected failure**; no skipped tests.
- [x] Focused WBGT tests: **208 passed** with Python warnings treated as errors.
- [x] Ruff lint/format and strict Sphinx HTML build passed.

Reproduce with Python 3.10:

```bash
python -m pip install 'tox<4' -e '.[dev,lwbgt]' -r docs/requirements.txt
tox -e py310
ruff check pythermalcomfort tests
ruff format --check pythermalcomfort tests
sphinx-build -W --keep-going -b html docs dist/docs
```

The corrected URLs were verified; full external link checking remains incomplete due to GitHub HTTP 429 responses.

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

# Reviewer's checklist

- [ ] The code is well-documented
- [ ] The code is well-tested
- [ ] The code follows the style guidelines of this project
- [ ] The code has been reviewed by a team member
- [ ] The code has been tested by a team member

## Summary by CodeRabbit

- **New Features**
  - Added the Liljegren outdoor Wet Bulb Globe Temperature (WBGT) model for estimating WBGT and component temperatures from meteorological inputs.
  - Supports scalar and array inputs, broadcasting, missing-data handling, diagnostics, configurable dates, averaging, wind height and output rounding.
  - Added clear status reporting for successful, partial and failed calculations.

- **Documentation**
  - Added installation guidance for the optional `lwbgt` extra, model usage details and references.
  - Updated links to the contributing guide.

- **Chores**
  - Included the optional backend in development and test environments.


FedericoTartarini:
Thank you so much for working on this, however I have a few questions because this request is not clear to me. Why do we need to install this function using a specific command? as usual in the description of the request. Why do you call this function an API?

> this API performs the underlying calculations through optional [lwbgt>=0.3.0,<0.4](https://pypi.org/project/lwbgt/), installed with pythermalcomfort[lwbgt]

The following description is not clear to me and it seems to be written completely by AI.

> Includes array broadcasting, input validation, missing-data handling, solver diagnostics, tests and documentation. Assumptions and active wind adjustment generate one combined INFO message per call; unused wind settings are unlogged at 2 m.

docs/installation.rst
```diff
@@ -86,6 +86,56 @@ Development extras (used by maintainers and contributors):
 
 If you installed via ``pip install -e .[dev]`` these will be installed for you.
 
+Liljegren WBGT estimation requires the optional ``lwbgt`` extra:
```
FedericoTartarini, line 89: Why do we need to install an additional dependencies? I'm not very inclined of adding new dependencies because it makes the code much harder to maintain. What is this dependency doing?

pythermalcomfort/models/wbgt_liljegren.py
```diff
@@ -0,0 +1,308 @@
...
+
+import numpy as np
+
+from pythermalcomfort.classes_input import NumericInput, WBGTLiljegrenInputs
+from pythermalcomfort.classes_return import WBGTLiljegren
+
+_logger = logging.getLogger(__name__)
+
+
+def wbgt_liljegren(
```
FedericoTartarini, line 13: Why is you just adding the function is not sufficient? I think we should be just adding the function itself and not adding a new external dependency.

pythermalcomfort/classes_input.py
```diff
@@ -36,19 +36,27 @@ class BaseInputs:
     age: NumericInput = numeric_field()
     airspeed_control: bool = field(default=True, metadata={"is_bool": True})
     asw: NumericInput = numeric_field()
+    averaging_minutes: NumericInput = numeric_field()
```
FedericoTartarini, line 39: We also need to think about adding this new variable, this is only used by this specific model

setup.py
```diff
@@ -115,6 +115,7 @@ def read(*names: str, encoding: str = "utf8") -> str:
             "pandas",
             "seaborn>=0.13",
         ],
+        "lwbgt": ["lwbgt>=0.3.0,<0.4"],
```
FedericoTartarini, line 118: See my previous comment about installing dependencies and also is not clear why we need to specify such a narrow range of versions.

FedericoTartarini (changes requested):
Thank you very much for working on this. We should carefully consider whether adding this new dependency is necessary, as it could impact the maintainability of the codebase. Before moving forward, let’s pause and ask: what’s the actual value of this dependency? Can we instead provide the new function directly and let users pass inputs in the same way they do for all other functions? This would keep things simpler, reduce coupling, and make the code easier to maintain.

pythermalcomfort/classes_input.py
zyf0717: Thanks for bringing this up; in fact it highlighted a separate issue for lwbgt itself: this should be completedly coupled to the datetime fields. Will resolve accordingly.

setup.py
zyf0717: Happy to relax it here, since the lwbgt interface is intentionally stable and will remain so.

zyf0717:
> Thank you very much for working on this. We should carefully consider whether adding this new dependency is necessary, as it could impact the maintainability of the codebase. Before moving forward, let’s pause and ask: what’s the actual value of this dependency? Can we instead provide the new function directly and let users pass inputs in the same way they do for all other functions? This would keep things simpler, reduce coupling, and make the code easier to maintain.

Thanks for reviewing!

My intention with `lwbgt` was to provide a small, stable interface to the non-trivial Liljegren C kernel. Existing reimplementations can differ quite a bit in numerical behaviour, correctness and performance, partly because the model is non-trivial and implementations often add their own preprocessing and assumptions.

Architecturally, the intention for `lwbgt` as a package was to keep that complexity isolated, with bit-identical output vs. the original Argonne C implementation, and precompile wheels for the major supported platforms, rather than requiring `pythermalcomfort` or any other packages to maintain and build another implementation itself.

For reference, even a Python-native implementation will not be a one function: `thermofeel` has a ~300-line Liljegren solver module, and still expects solar geometry to be supplied externally (it recommends `earthkit-meteo`). Other Python packages require [compiling C](https://github.com/kwodzicki/pywbgt) or [Cython](https://github.com/QINQINKONG/PyWBGT).

That said, I should have made this clearer upfront: the dependency is intended to keep the integration simple. I’m happy to explore a native implementation as well, though I’m not yet convinced it can stay as simple while matching the numerical behaviour and performance 😂

pythermalcomfort/models/wbgt_liljegren.py
zyf0717: See separate reply.

docs/installation.rst
zyf0717: See separate reply.

FedericoTartarini:
Okay thank you very much for letting me know and I was not aware about this issue I would then consider whether we should add this model or not to pythermalcomfort. If the implementation is too complex and if it is already included in thermofill, maybe we may consider not included in pythermalcomfort

Issue #416, opened 2026-09-18
Keep the Heat Index legend on one row in ThresholdPlot Example 7


Mars-Wu-bq:
**Threshold plot (threshold_plot.ipynb / threshold.py)**

In Example 7 of `threshold_plot.ipynb`, the Heat Index legend contains five risk-level entries, while `ThresholdPlot` uses a default maximum of four columns. This causes one entry to wrap onto a separate row unnecessarily.

For this specific example, setting `ncol=5` keeps all five entries on one row without changing the global legend defaults.

This task was originally listed in #338 and is being separated so it can be reviewed and submitted as a small, atomic pull request.

Mars-Wu-bq:
## Proposed approach

Set the Example 7 Heat Index legend to five columns so all five risk levels remain on one row:

```python
legend_kws={"title": "NWS risk level", "ncol": 5}
```

This change would apply only to this notebook example and would not affect the default `ThresholdPlot` legend behavior. @FedericoTartarini 

<img width="2683" height="993" alt="Image" src="https://github.com/user-attachments/assets/fff3b21e-c3e2-406c-be51-be1cd16bb834" />

FedericoTartarini:
Yes, please go ahead. Please note that all your images that you have added to the issues are broken and they are not displaying.

Mars-Wu-bq:
> Yes, please go ahead. Please note that all your images that you have added to the issues are broken and they are not displaying.

I have re-uploaded all the images and checked the page on another device to confirm that they are displaying correctly.

I occasionally encounter the same issue where the images do not load, but refreshing the page usually resolves it. If you are still unable to see the images, please let me know.

FedericoTartarini:
Thank you for sharing this example. I've noticed that the figure that you have pasted here in this issue looks different from the figure in the pull request.

Mars-Wu-bq:
> Thank you for sharing this example. I've noticed that the figure that you have pasted here in this issue looks different from the figure in the pull request.

The figure in the issue was generated about two weeks ago. When preparing the pull request, I found that my local branch was behind the latest `development` updates, so I updated it, reran the example, and generated a new figure. The title placement changed as a result, but the plotted data and the behavior demonstrated by the example remain the same.

stefanoschiavon:
@Mars-Wu-bq I suggest changing the legend title to legend_kws={"title": "US National Weather Service risk level", "ncol": 5}

PR #423, opened 2026-09-20
fix: keep heat index legend entries on one row

Mars-Wu-bq:
## Summary

ThresholdPlot Example 7 displays five NWS Heat Index risk levels, but the default four-column legend causes one entry to wrap onto a second row.

This PR sets `ncol=5` for this notebook example so all five entries remain on one row. The global `ThresholdPlot` legend defaults are unchanged.

This is a small change to a documentation example and does not affect the library API or calculation logic.

Fixes #416.

## Before and after

ThresholdPlot Heat Index legend before and after

## Testing

- `pipenv run ruff check docs/documentation/plots/matplotlib/threshold_plot.ipynb` — passed
- `pipenv run ruff format --check docs/documentation/plots/matplotlib/threshold_plot.ipynb` — passed
- The modified notebook is valid JSON.

## Summary by CodeRabbit

- **Documentation**
  - Updated the Heat Index example’s legend to display entries across three columns while retaining the “NWS risk level” title.


FedericoTartarini:
Thank you for offering to work on this feature. I’d like to suggest a few refinements to the chart design. After reviewing the update to five columns, I notice excess whitespace on the right side of the chart. This disrupts the visual balance. Ideally, the legend should be centered with the chart itself—this should be the default behavior. Alternatively, we could set it to three columns, which would provide three labels per row and likely improve overall readability and alignment.

Mars-Wu-bq:
> Thank you for offering to work on this feature. I’d like to suggest a few refinements to the chart design. After reviewing the update to five columns, I notice excess whitespace on the right side of the chart. This disrupts the visual balance. Ideally, the legend should be centered with the chart itself—this should be the default behavior. Alternatively, we could set it to three columns, which would provide three labels per row and likely improve overall readability and alignment.

Thank you for the feedback. I’ve updated the legend to use three columns. Please let me know if you would prefer any further changes.
image

FedericoTartarini (approved):

stefanoschiavon:
@Mars-Wu-bq and @FedericoTartarini, the HI chart is good, it does not have the black line on the top and right that determine the graph area and it does not have gridlines. Also here, it woudl be valuable to update the title to legend_kws={"title": "US National Weather Service risk level", "ncol": 5}

